PR #23, opened 2023-04-19
Read publishableKey from env for storybook

razor-x:
Updated Vercel config to set this env var.

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -4,7 +4,9 @@ import { SeamProvider } from 'index.js'
 
 // UPSTREAM: Use toolbar input when it lands.
 // https://github.com/storybookjs/storybook/pull/21959
-const publishableKey = 'seam_pk1fGd41X_zKs0ZELRWEc8nWxiBsrTFC98'
+const defaultPublishableKey =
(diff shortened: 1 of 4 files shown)


.storybook/preview.tsx
```diff
@@ -1,10 +1,12 @@
...
+import { type Preview } from '@storybook/react'
 
 import { SeamProvider } from 'index.js'
 
 // UPSTREAM: Use toolbar input when it lands.
 // https://github.com/storybookjs/storybook/pull/21959
-const publishableKey = 'seam_pk1fGd41X_zKs0ZELRWEc8nWxiBsrTFC98'
+const defaultPublishableKey =
+  process.env['STORYBOOK_SEAM_PUBLISHABLE_KEY'] ??
+  'seam_pk1fGd41X_zKs0ZELRWEc8nWxiBsrTFC98'
```
rchodava, line 9: what's the default here when env var not set? should we just error instead if not set?

rchodava (approved):

.storybook/preview.tsx
razor-x: This is for zero-setup local development. The env var is used as an override option or in deployment. Eventually there will be in input box in storybook for it.

PR #43, opened 2023-04-21
Normalize type-only imports

razor-x:
If an import only imports types, the type keyword should be on the outside or it will leave an empty import.

I have not found a linter rule that handles this correctly. I think it may be a known upstream issue.


razor-x:
> @razor-x we have this linter rule in some other repos, no? I think I saw this warning/error?

The rules I am aware of will catch type only imports without the `type` keyword, but I don't think they catch the keyword on the inside vs. outside. The rule we are looking for is one that enforces the `type` keyword on the outside if all imports are type-only, and inside otherwise. If we can find that rule let's add it!

PR #45, opened 2023-04-25
Use SEAM_ENDPOINT if defined in docs

razor-x:
Will use the real Seam endpoint (no proxy) on deployment. I suspect we will still want the local dev proxy, but can test that after this is confirmed working.


seveibar (approved):

razor-x:
CORS setup on the API is updated and this is now working!

PR #47, opened 2023-04-26
Initial Components: Device Details & Access Code Table

mikewuu:
Moved lots, renamed lots. Not saying we have to keep the changes, but it was just easier to show most of it rather than typing out, so here it is. I suppose this PR is more of an RFC, it's meant to be taken apart, so we can hopefully land on some consistent conventions going forward.

Style guide mostly follows the [Airbnb style guide](https://airbnb.io/javascript/react/).

Run `npm run start` to see the components in storybook

### Screenshots

CleanShot 2023-04-26 at 13 59 38

### Other Notable things

- Removed `/hooks/*`. Personally I find it easier to find things when they're named by use-case (vs. type). Here being able to browse the seam api by directory, and seeing the resources available is helpful.
- Turned off the `explicit-return` rule to let TS infer away.
- We don't have webpack! Not sure if a module bundler is incoming, but in the meantime I've added a  manual sass compile script, and used svg strings for images. 
- Don't have a watcher for scss file changes, so you'll need to re-run `npm run build` each time to have it take effect.
- Not sure why TS is not happy with MUI components, could use a hand here: 

**TS Error for MUI components:** 

```
JSX element type 'Dialog' does not have any construct or call signatures.
```


.gitignore
```diff
@@ -205,7 +206,8 @@ $RECYCLE.BIN/
 .LSOverride
 
 # Icon must end with two \r
-Icon
```
razor-x, line 208: This is actually important. Can you revert this?

.gitignore
```diff
@@ -22,6 +22,7 @@ docs
 
 # Build directories
 package
+dist
```
razor-x, line 25: Hmm, what is building to dist?

src/lib/SeamComponents.tsx
```diff
@@ -1,5 +1,11 @@
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
```
razor-x, line 1: Was there a reason for this rename? I think we should keep the name `SeamProvider`. The file rename to match the component name is ok though.

package.json
```diff
@@ -70,13 +71,15 @@
...
     "@tanstack/react-query": "^4.28.0",
-    "seamapi": "^5.15.0",
+    "seamapi": "^5.17.0",
     "uuid": "^9.0.0"
   },
   "devDependencies": {
     "@babel/preset-env": "^7.21.4",
     "@babel/preset-react": "^7.18.6",
     "@babel/preset-typescript": "^7.21.4",
+    "@emotion/styled": "^11.10.6",
```
razor-x, line 81: If we are using `emotion` and `material` do we need those as dependencies? And if we are brining those in, why have two style system (emotion and scss)?

src/lib/SeamComponents.tsx
```diff
@@ -100,3 +108,13 @@ const createSeamContextValue = ({
 const defaultSeamContextValue = createDefaultSeamContextValue()
 
 export const seamContext = createContext<SeamContext>(defaultSeamContextValue)
+
+export function useSeamComponents() {
```
razor-x, line 112: If we revert the rename.
```suggestion
export function useSeamContext() {
```

razor-x, line 112: The original name was because it was a private hook. Why do we need this outside of this module (should it be part of the public API)?

.eslintrc.json
```diff
@@ -20,7 +20,8 @@
         "fixStyle": "inline-type-imports"
       }
     ],
-    "@typescript-eslint/no-unused-vars": "off",
+    "@typescript-eslint/no-unused-vars": "warn",
+    "@typescript-eslint/explicit-function-return-type": "off",
```
razor-x, line 24: Yea, it's a bit much, and I found it annoying at first, but it grew on me. It speeds up TSC and really forces us to define our API. Counter arguments?

.eslintrc.json
```diff
@@ -20,7 +20,8 @@
         "fixStyle": "inline-type-imports"
       }
     ],
-    "@typescript-eslint/no-unused-vars": "off",
+    "@typescript-eslint/no-unused-vars": "warn",
```
razor-x, line 23: This is intentional, it's handled by the import plugin so we have to turn this one off.

```suggestion
```

.storybook/preview-head.html
```diff
@@ -0,0 +1,4 @@
+<link
```
razor-x, line 1: If we are doing this, it means this has be in the docs in the README as a requirement to use the components.

.storybook/preview.tsx
```diff
@@ -1,6 +1,8 @@
 import type { Preview } from '@storybook/react'
 
-import { SeamProvider } from 'index.js'
+import '../lib/styles/index.css'
```
razor-x, line 3: We will need to document how to import the styles with our steps to using the lib.

build-css.cjs
```diff
@@ -0,0 +1,7 @@
+const sass = require('sass')
```
razor-x, line 1: We will find a proper solution for this (maybe swap tsc with tsup?), but I'm ok with this hack for the moment. Discussion here: https://github.com/seamapi/react/discussions/36

src/lib/icons/access-code-key.ts
```diff
@@ -0,0 +1,4 @@
+export const accessCodeKey = `<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
razor-x, line 1: We should make the bundler hanlde svg => react component. Let's add that to the discussion: https://github.com/seamapi/react/discussions/36

src/lib/seam/client.ts
razor-x, line 1: Can we name this `use-seam-client.ts`?

src/lib/seam/client.ts
```diff
@@ -1,18 +1,17 @@
 import { useQuery } from '@tanstack/react-query'
-import { useContext } from 'react'
 import { type ClientSession, Seam } from 'seamapi'
 import { v4 as uuidv4 } from 'uuid'
 
-import { seamContext } from 'lib/provider.js'
+import { useSeamComponents } from 'lib/SeamComponents.js'
 
-export function useSeam(): {
+export function useSeamClient(): {
```
razor-x, line 7: I left this as `useSeam` intentionally (considered `useSeamClient`) but I wanted to leave the option open to put more in the context. Not I'm not so sure and think this change is probably correct!

src/lib/components/index.ts
razor-x, line 1: Let's update the codeowner file to match now that this directory is no more!

src/lib/ui/TextField/index.tsx
razor-x, line 1: I really would like to avoid the pattern of having code in `index.*` files. Let's name these like `TextField/TextField.tsx`. Same for the others.

src/lib/ui/DeviceManager/index.tsx
```diff
@@ -1,14 +1,14 @@
...
-import { useDevices, type UseDevicesParams } from 'lib/hooks/use-devices.js'
+import { useDevices, type UseDevicesParams } from 'lib/seam/devices/list.js'
 
 export type DeviceManagerProps = UseDevicesParams
 
 /**
  * Fetch, list, and manage devices.
  */
-export const DeviceManager = (props: DeviceManagerProps): ReactElement => {
+export default function DeviceManager(props: DeviceManagerProps) {
```
razor-x, line 11: I'm not sure we need to export default exports. What is the use case? Everything will be imported by named export right? I think we should keep `export const DeviceManager` style in that case.

lib/styles/index.css
razor-x, line 1: Is this generated? We should not check this in to git. Let's just add it to gitignore for now and add it to the files array in package.json.

razor-x (changes requested):
Getting late here, so please excuse any typos or terse sounding comments. Mostly just trying to go-fast b/c such PR, much big. This is a great start! Some of the structure changes are great, but have some concerns about others. It's important we sort those out now as it will get harder to change our mind later about some of these conventions.

razor-x:
Side note on the CSS stuff: even without a bundler for `npm:build` at the moment, we still need to make sure it can integrate into our dev flow for storybook / the examples which do have proper tooling.

razor-x:
>  I suppose this PR is more of an RFC, it's meant to be taken apart, so we can hopefully land on some consistent conventions going forward.

Missed this on my first look, but +1 to this approach!

.storybook/preview.tsx
mikewuu: Added simple import for now in  https://github.com/seamapi/react/pull/47/commits/da4137d6ad849297e4dcdeb9a8eacb6501e49492

mikewuu: We should definitely circle back, and update docs once everything is more production-ready.

build-css.cjs
mikewuu: Exactly. This was definitely just to get something out the door vs. messing with webpack configs for 3 days.

package.json
mikewuu: Only planning to use `material` (`emotion` is a MUI dep) in the storybook demos. Not the final bundle.

src/lib/SeamComponents.tsx
mikewuu: Not really, just wanted to get some thoughts. Updated in https://github.com/seamapi/react/pull/47/commits/cf2023f528bba35db575bd5a17b3a48dd56c0fc1

src/lib/SeamComponents.tsx
mikewuu: This hook just combines `useContext` with `seamContext`. `seamContext` was already previously exported (and `useContext`) was called separately.

I don't think there are any significant changes in terms of exposed API. Co-locating with the provider just makes it a little easier to use, and handles the case where it's called outside `<SeamProvider/>` by throwing an error.

src/lib/seam/client.ts
mikewuu: I'm not a fan of `use-*` file names because it results in having 1 hook per file, and that can result in way too many files (or mixing conventions). I can see the appeal though, it does make searching for the hook easier.

Open to other takes here, @azat-co  @seveibar @xplato ?

src/lib/seam/client.ts
mikewuu: This is specifically for the API client, though. We still have `useSeamContext` in `lib/SeamProvider`

src/lib/ui/DeviceManager/index.tsx
mikewuu: I'm used to the primary component being a default export. This is inline with how MUI does things too.

That said, if we want to avoid default exports, that's fine too, it can just be:

```ts
export function DeviceManager
```

I would really prefer to keep components as named `function`(s); however, as in JS, only functions are hoisted. That is, using `const` would force us to put dependencies at the top.

I'm used to writing (and reading) code top-down with each dependency coming after the above. IMO it helps readability more than using the shorter `const` for definitions.

mikewuu:
Hmm, storybook shows the new story locally, but it's missing in the preview build.

https://react-ad8qlxqm1-seamapi.vercel.app/?path=/docs/seam-react-components--docs

@razor-x , @seveibar  any ideas here?

.gitignore
razor-x: Try `git checkout main .gitignore`

src/lib/ui/DeviceManager/index.tsx
razor-x: > That said, if we want to avoid default exports

Yes, I would like to essentially prohibit all default exports from the project, expect possibly on a special case-by-case bases. Probably for things that looks like assets, like the SVG components, css, images, "json" files written in TS, etc. In those cases, the export would not be named.

When default / named exports are mixed it leads to inconsistency and can break treeshaking.

> I would really prefer to keep components as named function(s); however, as in JS, only functions are hoisted. That is, using const would force us to put dependencies at the top.

We can standardize on named functions, but I don't think the hoisting argument is correct. `const` and `let` are hoisted and deps can be defined below the component that uses them, see here: https://github.com/seamapi/react/blob/main/src/lib/components/DeviceManager/DeviceManager.tsx#L11

Can you provide an example of what `function` lets us do that `const` does not? Otherwise it's just a style preference between arrow or function (can have that debate separately if that's the case).

💯 For deps below the main component though.

.gitignore
razor-x: Maybe, but let's decide that in the bundler discussion.

src/lib/SeamComponents.tsx
razor-x: Ok, I think SeamProvider is more ubiquitous: people instantly recognize the provider pattern that way.

package.json
razor-x: Ooo, interesting, so like UI to control the components?

src/lib/seam/client.ts
seveibar: i like the `use-*` prefix, makes developers less creative with file names

seveibar (approved):

src/lib/seam/client.ts
razor-x: Let's stick with `use-*`, but if we find a case where we need multiple hooks or exports in a single file, we can consider the exception when we have the need.

src/lib/ui/AccessCodeTable/AccessCodeTable.stories.tsx
```diff
@@ -1,6 +1,4 @@
-import MuiContainer from '@mui/material/Container'
-import Dialog from '@mui/material/Dialog'
-import DialogContent from '@mui/material/DialogContent'
+import { Container as MuiContainer, Dialog, DialogContent } from '@mui/material'
```
razor-x, line 1: @mikewuu Just want to note that we should always import like this from MUI everywhere (and for any libs, prefer named vs default if possible). That will be the friendliest for the bundler and treeshaking, and groups everything together under a single import.

README.md
```diff
@@ -50,6 +50,21 @@ export const App = () => {
 
 [Seam Console]: https://console.seam.co/
 
+### Fonts
+
+The components have been designed with Source Sans Pro in mind with fallback to other system sans-serif fonts. To use it with Source Sans Pro, please load the font from Google Fonts by placing the following in you `<head>` tag:
```
razor-x, line 55:
```suggestion
The components are optimized for use with [Source Sans Pro](https://fonts.google.com/specimen/Source+Sans+Pro), but will fallback to other system sans-serif fonts.
Load the font from Google Fonts by placing the following in you `<head>` tag:
```

src/lib/ui/DeviceManager/index.tsx
mikewuu: > Can you provide an example of what function lets us do that const does not?

Ya, but it's super contrived: https://codesandbox.io/s/react-playground-forked-gpl5fj?file=/index.js

https://user-images.githubusercontent.com/11449462/234869020-1dd9087e-f73a-4b3e-8b83-162f8a775085.mp4

In practice (with a bundler), you're right, it'd be _mostly_ a style choice since the bundler should move everything around, so it's all defined when it's used anyway. I say mostly because I have run into `access before init` errors that went away after changing `const` -> `function`. May have been due to the project's dependency graph though. 🤷‍♂️

package.json
mikewuu: Yea, just as examples.

src/lib/ui/AccessCodeTable/AccessCodeTable.stories.tsx
mikewuu: Yea! This fixed the weird not callable error. 

I usually prefer default imports though because I find named imports slowed down TS:

CleanShot 2023-04-27 at 22 26 45

https://mui.com/material-ui/guides/minimizing-bundle-size/#development-environment

.gitignore
mikewuu: Totally forgot you could do this, thanks!

reverted in https://github.com/seamapi/react/pull/47/commits/ec209b813c7886dddcb391d4a8ac9777ff7acb9c

src/lib/seam/client.ts
mikewuu: Renamed in https://github.com/seamapi/react/pull/47/commits/6650a9986aab5052a494f8fde2f1a048d3567ca8

src/lib/components/index.ts
mikewuu: Renamed in https://github.com/seamapi/react/pull/47/commits

lib/styles/index.css
mikewuu: Removed this, but there's already `lib` in `files`, so we should be good there, right?

package.json
```diff
@@ -35,16 +35,20 @@
     "!src/stories"
   ],
   "scripts": {
-    "start": "storybook dev --port 6006",
-    "build": "npm run build:ts",
+    "start": "concurrently \"npm run storybook:start\" \"npm run css:start\"",
```
mikewuu, line 38: > Side note on the CSS stuff: even without a bundler for `npm:build` at the moment, we still need to make sure it can integrate into our dev flow for storybook / the examples which do have proper tooling.

You're right, manually building css on changes was lame. Updated it here to auto build on css changes too:

https://user-images.githubusercontent.com/11449462/234885086-46895826-1293-4cb6-9542-f04bebdb1ec2.mp4

package.json
razor-x: I may move the style toolchain to it's own PR. We should probably integrate this https://storybook.js.org/recipes/@emotion/styled

src/lib/ui/DeviceManager/index.tsx
mikewuu: Removed default exports in https://github.com/seamapi/react/pull/47/commits

src/lib/ui/TextField/index.tsx
mikewuu: Removed in https://github.com/seamapi/react/pull/47/commits/0cd5b36d372c6800d62d86ab611c39f78bffbdb8

mikewuu:
@razor-x  I think I've addressed all the threads. Open items should also have been spun off into separate issues. If everything looks good I'd like to merge this PR first before opening another one to flesh out the rest of the UI.

src/lib/styles/reset.scss
razor-x, line 1: @mikewuu Is this reset taken from somewhere? I'd like to use or attribute the source.

razor-x: My Google searching popped up this as the most recent reset, but maybe there is a newer one? https://github.com/elad2412/the-new-css-reset

src/lib/ui/types.ts
```diff
@@ -0,0 +1,2 @@
+export type DivProps = React.HTMLAttributes<HTMLDivElement>
```
razor-x, line 1:
```ts
import type { HTMLAttributes } from 'react'
```

dawnho:
Overall look and feel looks good. Guessing that the menu / search / code count values will be hooked up later?

dawnho (approved):

src/lib/styles/reset.scss
mikewuu: Updated to use new css reset (from main)

.eslintrc.json
```diff
@@ -22,6 +22,8 @@
       }
     ],
     "@typescript-eslint/no-unused-vars": "off",
+    "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/restrict-plus-operands": "off",
```
mikewuu, line 26: Was complaining about `Date.now() + ms('1d')` even though both are numbers. 🤷‍♂️

tsconfig.json
```diff
@@ -14,7 +14,7 @@
     "verbatimModuleSyntax": true,
     "allowUnreachableCode": false,
     "allowUnusedLabels": false,
-    "exactOptionalPropertyTypes": true,
+    "exactOptionalPropertyTypes": false,
```
mikewuu, line 17: This was a super weird one. It would fail optional component props. Things like `onClick?: () => void`

.stylelintrc.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": ["stylelint-config-standard-scss"]
+  "extends": ["stylelint-config-standard-scss"],
+  "rules": {
+    "no-invalid-position-at-import-rule": null
```
mikewuu, line 4: We're nesting `@import` in `.seam-components`, so I'm not sure this is required?

package.json
```diff
@@ -91,13 +91,18 @@
...
     "seamapi": "^5.19.0",
     "uuid": "^9.0.0"
   },
   "devDependencies": {
     "@babel/preset-env": "^7.21.4",
     "@babel/preset-react": "^7.18.6",
     "@babel/preset-typescript": "^7.21.4",
+    "@emotion/styled": "^11.10.6",
+    "@mui/icons-material": "^5.11.16",
+    "@mui/material": "^5.12.2",
```
seveibar, line 105: fine for now based on usage

tsconfig.json
seveibar: agree

seveibar (approved):
lgtm!

package.json
mikewuu: These are only to quickly scaffold out examples in the stories. Won't be using them in the actual components, and they won't be included in the consumer's final bundle.

razor-x (approved):
Agreed on early merge with @seveibar and will have a follow up PR instead of blocking review.

src/index.scss
```diff
@@ -1,2 +1 @@
-@use 'the-new-css-reset/css/reset.css';
-@use './styles/_main.scss';
+@use './styles/main.scss';
```
razor-x, line 1: @mikewuu Should we follow this convention? https://sass-lang.com/guide

> Partials
> You can create partial Sass files that contain little snippets of CSS that you can include in other Sass files. This is a great way to modularize your CSS and help keep things easier to maintain. A partial is a Sass file named with a leading underscore. You might name it something like _partial.scss. The underscore lets Sass know that the file is only a partial file and that it should not be generated into a CSS file. Sass partials are used with the @use rule.

.stylelintrc.json
razor-x: I wonder if we can use `@use` instead and make writing all the styles a bit easier. Docs suggest (1) `@import` is going to be removed and (2) `@use` can namespace so we don't need to prefix all the styles, Sass will do that for us.

razor-x: https://github.com/seamapi/react/discussions/81

src/lib/use-toggle.ts
```diff
@@ -0,0 +1,18 @@
+import { useCallback, useState } from 'react'
+
+/**
+ * Simple util to quickly handle on/off states.
+ */
+const useToggle = (initialState: boolean = false): [boolean, () => void] => {
+  const [isTrue, setIsTrue] = useState(initialState)
+
+  // use callback otherwise it would yield unexpected results in promises
```
razor-x, line 9: Interesting, curious how you learned about this. Is it specific to `useToggle` or is this something you always need to worry about with `useState`?

src/lib/ui/AccessCodeTable/AccessCodeTable.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+export const Content: Story = {}
+
+export const InsideModal: Story = {
+  render: (props) => {
+    // Wrap modal/dialog contents in `seam-components` class
+    // to apply styles when rendered in a portal,
+    // which is the the default MUI behavio.
+    return (
+      <Dialog open fullWidth maxWidth='sm'>
+        <div className='seam-components'>
```
razor-x, line 91: Will this make integration difficult for our consumers? Do we need to document this or can we find a solution that will wrap the model in the right class?

src/lib/ui/AccessCodeTable/CodeDetails.tsx
```diff
@@ -0,0 +1,51 @@
...
+import type { AccessCode } from 'seamapi'
+
+import { DotDivider } from 'lib/ui/layout/DotDivider.js'
+
+export function CodeDetails(props: { accessCode: AccessCode }) {
+  const { accessCode } = props
+
+  return (
+    <div className='seam-access-code-details'>
+      Unit 110
```
razor-x, line 11: Where do we need to load this from?

mikewuu: That's the device name, no hooks implemented yet at the moment.

src/lib/ui/AccessCodeTable/AccessCodeTable.stories.tsx
mikewuu: > Will this make integration difficult for our consumers? 

Yea, it does assume the consumer knows the difference between rendering in the DOM tree, and via portal somewhere else. IMO the example code should act as sufficient documentation?

mikewuu: > will wrap the model in the right class?

Interesting... we _could_ just add the class to all the modal contents in case they're rendered somewhere else. Having duplicate parent classes of `.seam-components` shouldn't pose issues.

src/index.scss
mikewuu: Yep, I did try this yesterday, but ran into a couple roadblocks. Will spin off into a new issue to refactor to use `@use`

mikewuu: https://github.com/seamapi/react/issues/84

src/lib/ui/AccessCodeTable/AccessCodeTable.stories.tsx
mikewuu: https://github.com/seamapi/react/issues/85

.eslintrc.json
mikewuu: Oh? May have been my vscode. It was definitely weird, I was seeing both as `number` yet it wouldn't add!

mikewuu: https://github.com/seamapi/react/issues/86

PR #49, opened 2023-04-26
Set 8gb devcontainer requirement

razor-x:
Webpack hungry. Feed webpack more ram!


seveibar:
fwiw the m1s some devs at Seam (myself included) have only have 8gb ram total iirc. I think this is still ok?

seveibar (approved):

razor-x:
Yea, I think this is more of a devcontainer thing than local issue

PR #61, opened 2023-04-28
Strict CSP

razor-x:
Ensure we support and document how to use the components with CSP.


razor-x:
Use https://www.npmjs.com/package/webpack-plugin-replace and also https://webpack.js.org/plugins/copy-webpack-plugin/ over `cpy`.

mikewuu (approved):

PR #62, opened 2023-04-28
Sort devices by id

razor-x:
Just a quick fix to avoid the chromaic UI diff on each PR

### DIFF
--- a/src/lib/components/DeviceManager/DeviceManager.tsx
+++ b/src/lib/components/DeviceManager/DeviceManager.tsx
@@ -16,9 +16,11 @@ export const DeviceManager = (props: DeviceManagerProps): ReactElement => {
 
   return (
     <ul>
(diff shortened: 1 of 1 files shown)


src/lib/components/DeviceManager/DeviceManager.tsx
```diff
@@ -16,9 +16,11 @@ export const DeviceManager = (props: DeviceManagerProps): ReactElement => {
 
   return (
     <ul>
-      {devices?.map((device) => (
-        <DeviceManagerItem key={device.device_id} {...device} />
-      ))}
+      {devices
+        ?.sort((a, b) => (a.device_id > b.device_id ? 1 : -1))
```
seveibar, line 20: hmm created_at might be a better default

seveibar (approved):

PR #46, opened 2023-04-26
Use seam-connect-fake for local development

razor-x:
- Runs and uses the [fake-seam-connect](https://github.com/seamapi/fake-seam-connect) for the local storybook.
- Uses fake for Jest (almost working, try here for exit code fix https://github.com/open-cli-tools/concurrently/issues/187#issuecomment-1274659502)

PR #74, opened 2023-05-03
Integrate Seam Connect Fake + Basic Hooks

seveibar:
- feat: Use seam-connect-fake for local storybook
- test: Use seam-connect-fake in test


razor-x:
@seveibar I went down the concurrently route and it has issues. This PR is the one we want, just need to resolve the merge conflicts https://github.com/seamapi/react/pull/46

PR #75, opened 2023-05-03
Seam Connect Fake Integration + Seed

seveibar:
Introduces Seam Connect Fake jest tests and storybook integration

The jest testing still needs to be thought out a bit more, but the storybook runs against the fake smoothly which is a good place to begin.


razor-x:
@seveibar Looks like you might have a global `.npmrc` that is overriding the `@seamapi` namespace to resolve to GitHub Packages instead of npm, which would explain why the entry in the lockfile points there now. If you remove that override and re-install the package it should fix the build.

.gitignore
```diff
@@ -210,7 +210,8 @@ $RECYCLE.BIN/
 .LSOverride
 
 # Icon must end with two \r
-Icon
```
razor-x, line 213: The `^M` is actually correct.

.npmrc
```diff
@@ -0,0 +1,2 @@
+registry=https://registry.npmjs.org/
```
razor-x, line 1: This file is in `.gitignore`. If we add it we should comment that line out. I'm not sure we should add this though, it's not a standard pattern I've seen.

.storybook/middleware.js
```diff
@@ -1,13 +1,20 @@
...
+  const fake = await create()
+
+  seedFake(fake.database.getState())
+
+  await fake.startServer()
   app.use(
-    '/',
-    createProxyMiddleware('/api', {
-      target: 'https://connect.getseam.com',
+    '/api',
```
razor-x, line 15:
```suggestion
    '/',
```

.storybook/preview.tsx
```diff
@@ -21,9 +20,13 @@ const preview: Preview = {
   },
   decorators: [
     (Story) => {
+      // TODO: This should be given by the story, some stories will want
```
razor-x, line 23: Can move to issue 👍🏻

jest.config.cjs
```diff
@@ -1,20 +1,34 @@
...
+    globalSetup: '<rootDir>/test/jest/global-setup.cjs',
+    globalTeardown: '<rootDir>/test/jest/global-teardown.cjs',
+    setupFilesAfterEnv: ['<rootDir>/test/jest/setup-api.ts'],
+    preset: 'ts-jest/presets/default-esm',
+    testEnvironment: 'jsdom',
+    moduleNameMapper: {
+      '^(\\.{1,2}/.*)\\.js$': '$1',
+      '^fixtures/(.*).js$': '<rootDir>/test/fixtures/$1',
+      '^lib/(.*).js$': '<rootDir>/src/lib/$1',
+      '^index.js$': '<rootDir>/src/lib',
```
razor-x, line 21:
```suggestion
```

jest.config.cjs
```diff
@@ -1,20 +1,34 @@
...
+      '^@seamapi/react': '<rootDir>/src/lib',
```
razor-x, line 22: This was originally supposed to be like this, suggesting here to reduce merge conflicts.
```suggestion
      '^@seamapi/react': '<rootDir>/src',
```

.npmrc
```diff
@@ -0,0 +1,2 @@
...
+@seamapi:registry=https://registry.npmjs.org/
```
razor-x, line 2:
```suggestion
@seamapi:registry=https://registry.npmjs.org/

```

seveibar:
@razor-x chromatic needs some env vars and in general a completely different, local configuration- this should not be holding us up at this point. Let's just disable this.

image

test/fixtures/react.tsx
```diff
@@ -1,11 +1,31 @@
 import { SeamProvider } from '@seamapi/react'
 import { render } from '@testing-library/react'
-import type { PropsWithChildren } from 'react'
```
razor-x, line 3: Interesting, did this stop working somehow 🤔

test/jest/global-setup.cjs
```diff
@@ -0,0 +1,53 @@
...
+
+  const clientSession2 = db.addClientSession({
+    workspace_id: ws2.workspace_id,
+    connect_webview_ids: [cw.connect_webview_id],
+    connected_account_ids: [ca.connected_account_id],
+    user_identifier_key: 'seed_client_session_user_2',
+  })
+
+  await fake.startServer(url.port)
+  projectConfig.globals.JEST_SEAM_PUBLISHABLE_KEY_1 = ws1.publishable_key
```
razor-x, line 49: I thought I tried this and it didn't work. Glad that you got it working. I think this means would could set `projectConfig.globals.JEST_SEAM_ENDPOINT` here as well.

package.json
```diff
@@ -91,15 +91,16 @@
   },
   "dependencies": {
     "@tanstack/react-query": "^4.28.0",
+    "seamapi": "^5.19.0-alpha.2",
```
razor-x, line 94: Weird, something broke the sorting.

razor-x (approved):
Let's go! Just found some minor things, but non-blocking. Mostly commenting so I can remember fix or look at them them later if they are not updated here.

razor-x:
> chromatic needs some env vars and in general a completely different, local configuration- this should not be holding us up at this point. Let's just disable this.

Just taking a quick look, it giving this error. Did something change on seamapi SDK? It is trying to PUT now.

image

seveibar:
> Just taking a quick look, it giving this error. Did something change on seamapi SDK? It is trying to PUT now.

I was wrong, I thought it had the wrong client session, but this is because the fake has PUT support (in the alpha seamapi sdk https://github.com/seamapi/javascript/pull/193) but the production API doesn't support it yet. That's actually a lot better than Chromatic having a configuration issue- I had incorrectly assumed it was the hard-coded PK i changed- nice catch

razor-x:
All good! I patched the minor issues on my new PR here: https://github.com/seamapi/react/pull/83

For the PUT issue, let's update the internal endpoint to match :)

PR #76, opened 2023-05-03
SVG - update generate to return JSX.Element

mikewuu:
Currenly returning `ReactNode` which isn't a valid react element for function components. This PR updates the return type to `JSX.Element`

TS error:

CleanShot 2023-05-03 at 12 05 17@2x

Reference: https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement/72353143#72353143


razor-x (approved):
Hmm I've gotten this issue before and use ReactElement, but I wonder if JSX.Element should be the preferred return types in all cases 🤔

mikewuu:
IIRC either works if we're not specifying the props. `JSX.Element` is an alias for `ReactElement<any>`

PR #80, opened 2023-05-03
Initial Hooks

seveibar:
builds off #75 

image


seveibar:
@razor-x this PR is good locally as shown in screenshot, but breaks in prod because CSTs can't call /access_codes/list https://github.com/seamapi/seam-connect/issues/2896

.eslintrc.json
```diff
@@ -29,6 +29,8 @@
     ],
     "@typescript-eslint/no-unused-vars": "off",
     "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/restrict-plus-operands": "off",
```
razor-x, line 32: I don't think we need to disable this, but curious where you hit the issue.

.eslintrc.json
```diff
@@ -29,6 +29,8 @@
...
+    "@typescript-eslint/strict-boolean-expressions": "off",
```
razor-x, line 33: IMO the relatively recent `?.` and preferring `x == null` and `x != null` now provides a nice balance of ergonomics / explicit code, e.g., `if (x?.length > 0)` and `if (x?.f?.x != null)`

.gitignore
```diff
@@ -234,3 +234,4 @@ Icon
 Network Trash Folder
 Temporary Items
 .apdisk
+.vscode
```
razor-x, line 237: Fine in this PR, but just a reminder to myself to PR this change https://github.com/seamapi/fake-template/commit/c69e7da4a9466cbdb579db79138cb30f091bc134

src/lib/hooks/use-device.ts
```diff
@@ -0,0 +1,26 @@
+import { useQuery } from '@tanstack/react-query'
+import type { Device, SeamError } from 'seamapi'
+
+import { useSeamClient } from 'lib/seam/use-seam-client.js'
+import type { UseSeamQueryResult } from 'lib/types/use-seam-query-result.js'
+
+export function useAccessCodes(
```
razor-x, line 7: Copy paste typo?
```suggestion
export function useDevice(
```

src/lib/hooks/use-seam-client.ts
```diff
@@ -0,0 +1 @@
+export { useSeamClient } from 'lib/seam/use-seam-client.js'
```
razor-x, line 1: Hmm, why the duplication of export locations for this and `use-devices`?

src/lib/types/use-seam-query-result.ts
```diff
@@ -0,0 +1,7 @@
+import type { UseQueryResult } from '@tanstack/react-query'
```
razor-x, line 1: Neat!

src/lib/hooks/use-device.ts
```diff
@@ -0,0 +1,26 @@
...
+  params:
+    | {
+        device_id: string
+      }
+    | string
+): UseSeamQueryResult<'device', Device<any>> {
```
razor-x, line 13: Just want to compare with the alternative of using these types instead. Curious your thoughts on one vs. the other.

```
export type UseDevicesParams = DevicesListRequest
export type UseDevicesData = DevicesListResponse['devices']
```

src/lib/hooks/use-access-codes.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { AccessCode, SeamError } from 'seamapi'
+
+import { useSeamClient } from 'lib/seam/use-seam-client.js'
+import type { UseSeamQueryResult } from 'lib/types/use-seam-query-result.js'
+
+export function useAccessCodes(params: {
+  device_id: string
+}): UseSeamQueryResult<'accessCodes', AccessCode[]> {
+  const { client } = useSeamClient()
+  const { data, ...rest } = useQuery<any, SeamError>({
```
razor-x, line 11:
```suggestion
  const { data, ...rest } = useQuery<AccessCodesListResponse['access_codes'], SeamError>({
```

src/lib/hooks/use-device.ts
```diff
@@ -0,0 +1,26 @@
...
+  const objParams = typeof params === 'string' ? { device_id: params } : params
+  const { client } = useSeamClient()
+  const { data, ...rest } = useQuery<any, SeamError>({
```
razor-x, line 16:
```suggestion
  const { data, ...rest } = useQuery<DevicesGetResponse['device'], SeamError>({
```

test/fixtures/react.tsx
```diff
@@ -1,7 +1,10 @@
-import { SeamProvider } from '@seamapi/react'
+import 'isomorphic-fetch'
```
razor-x, line 1: Weird, I wonder what changed that now requires this import?

test/fixtures/react.tsx
```diff
@@ -1,7 +1,10 @@
...
+
 import { render } from '@testing-library/react'
 import type { PropsWithChildren } from 'react'
 
+import { SeamProvider } from 'lib/SeamProvider.js'
```
razor-x, line 6: Did something break that required a different import?

src/lib/hooks/use-device.ts
razor-x, line 13: If we don't use the other req/res types:
```suggestion
): UseSeamQueryResult<'device', Device<CommonDeviceProperties>> {
```

src/lib/ui/AccessCodeTable/AccessCodeTable.tsx
```diff
@@ -8,17 +6,23 @@ import {
...
+  deviceId: string
   onClickBack?: () => void
 }): JSX.Element {
-  const { accessCodes, onClickBack } = props
+  const { accessCodes } = useAccessCodes({
+    device_id: props.deviceId,
+  })
+  const { onClickBack } = props
+
+  if (!accessCodes) return <>{null}</>
```
razor-x, line 25: Interesting, I think this caused a type error with `JSX.Element`. Can also fix via `return null` and updating type to `JSX.Element | null`

src/lib/seam/use-seam-client.ts
```diff
@@ -11,8 +12,10 @@ export function useSeamClient(): {
   error: unknown
 } {
   const { client, clientOptions, publishableKey, ...context } = useSeamContext()
-  const [clientSession, setClientSession] = useClientSession()
-  const userIdentifierKey = useUserIdentifierKey(context.userIdentifierKey)
+  const [clientSession, setClientSession] = useState<ClientSession | null>(null)
```
razor-x, line 15: It's possible we may not even need `useState` here anymore since `useQuery` should manage the lifecycle for all our request data.

src/lib/seam/use-seam-client.ts
```diff
@@ -52,34 +55,24 @@ export function useSeamClient(): {
...
-  const cachedClientSession = globalThis?.localStorage?.getItem(localStorageKey)
-
-  return [
-    cachedClientSession != null ? JSON.parse(cachedClientSession) : null,
-    setClientSession,
-  ]
-}
-
-function useUserIdentifierKey(userIdentifierKey: string | undefined): string {
+/**
```
razor-x, line 58: Cool. Better than a comment IMO: print a warning to console as well so the devs will see this!

src/lib/seam/use-seam-client.ts
```diff
@@ -52,34 +55,24 @@ export function useSeamClient(): {
...
-  const key = globalThis?.localStorage?.getItem(localStorageKey) ?? uuidv4()
-  globalThis?.localStorage?.setItem(localStorageKey, key)
-  return key
+  const fingerprintValue =
+    globalThis?.localStorage?.getItem('seam_user_fingerprint') ??
+    `fingerprint_${uuidv4()}`
+
+  globalThis?.localStorage?.setItem('seam_user_fingerprint', fingerprintValue)
+
+  return fingerprintValue
```
razor-x, line 77:
```suggestion
  const fingerprint =
    globalThis?.localStorage?.getItem('seam_user_fingerprint') ??
    `fingerprint_${uuidv4()}`

  globalThis?.localStorage?.setItem('seam_user_fingerprint', fingerprint)

  return fingerprint
```

razor-x (approved):
🔥😍🔥
Most review comments are things that can be discussed, fixed, or looked into later. Check the `useAccessCodes` hook name typo though before merge!

src/lib/hooks/use-device.ts
seveibar: harder for the consumer to read/understand

test/fixtures/react.tsx
seveibar: yea no idea, `fetch` was suddenly not found

test/fixtures/react.tsx
seveibar: do we prefer @seamapi/react within the test suite? Seems like a good rule

seveibar:
@razor-x i'm going to fix some of these post-merge b/c I think people might need to use the hooks today.

Access codes should be fixed by this (hopefully): https://github.com/seamapi/seam-connect/pull/2900

src/lib/hooks/use-seam-client.ts
razor-x: Ok, so you would like to add an entrypoint accessible via `@seamapi/react/hooks`. Can do! https://github.com/seamapi/react/issues/101

src/lib/seam/use-seam-client.ts
razor-x: Updated in https://github.com/seamapi/react/pull/99

PR #83, opened 2023-05-03
Followup changes to initial component PR

razor-x:
Changes are isolated to individual commits:

- refactor: Remove restrict-plus-operands override
- refactor: Remove duplicate line from .gitignore
- refactor: Use luxon
- feat: Always prefer PropsWithChildren
- fix: Add missing return types
- refactor: Simplify null checks
- build: Disable icon index file
- fix: Move never to t object
- refactor: Rename BATTERY_LEVEL to batteryThreshold
- refactor; Simplify null checks
- fix: Fix lock status typo
- refactor: Rename FALLBACK_IMAGE_URL to fallbackImageUrl
- refactor: Fix typo in comment


package.json
```diff
@@ -91,9 +91,9 @@
   },
   "dependencies": {
     "@tanstack/react-query": "^4.28.0",
-    "seamapi": "^5.19.0-alpha.2",
     "classnames": "^2.3.2",
-    "dayjs": "^1.11.7",
+    "luxon": "^3.3.0",
```
mikewuu, line 95: Should we remove `dayjs`, and `ms` if we're using `luxon`?

mikewuu (approved):

mikewuu:
Closes #86 
Closes #85

package.json
mikewuu: Ahh, sorry I see it IS removed.

razor-x: We could use dayjs or Luxon. Main reason for adding it here was to see how it works as an alternative and start a discussion. If we are ok with Luxon though then no need to hash it out, but it would be my first choice.

PR #96, opened 2023-05-08
Use fake pk when examples run locally with docs

razor-x:


examples/vite.config.ts
```diff
@@ -60,7 +61,18 @@ export default defineConfig(async ({ command }) => {
   }
 })
 
-const setupEnv = async (endpoint: string, isBuild: boolean): Promise<void> => {
+const setupEnv = async (
```
razor-x, line 64: This function signature is a bit contrived. Open to simplifications.

src/lib/seam/use-seam-client.ts
```diff
@@ -44,15 +45,17 @@ export function useSeamClient(): {
   return { client: data ?? null, isLoading, isError, error }
 }
 
-/**
- * You'll almost always want to supply a user identifier key, but if you don't
- * we'll automatically pick a fingerprint for this user. The user interface
- * will show warnings when you use a fingerprint because this client session
- * is bound to this machine, and is effectively ephemeral.
- */
```
seveibar, line 52: Jsdocs also appear in type hints for consumers, not everyone is going to look at the warning panel

seveibar: Unrelated but github never committed your suggestion for fingerprintValue which is concerning and annoying

razor-x: Ah no worries about the fingerprintValue, we will resolve those open comments as we go.

I didn't know that about the comments showing up in the type hints. Can we have both? For the type hints though, I think the comment would need to be on something directly used by the consumer, and right now we don't have any jsdoc comments for the public API. If we did we could annotate the `userIdentifierKey` prop with the warning.

Issue #108, opened 2023-05-09
Access Code Table - Update story to use `<AccessCodeTable/>` directly


mikewuu:
Currently the table accepts static `accessCodes`, this task is to update the component to use the actual hooks / preferred data-bindings.

mikewuu:
As a part of this task, once the fake is ready, we should render `AccessCodeTable` directly within **AccessCodeTable.stories** . We're currently rendering the Content component directly as a work-around.

mikewuu:
Technically the original task is already one in #80 , updated title to reflect the remaining update required once the fake is ready for local / storybook.

Issue #107, opened 2023-05-09
Access Code Table - Add ability to copy code from action menu


mikewuu:
Users can click the 'more actions' menu, and copy the code:

![Image](https://user-images.githubusercontent.com/11449462/236983589-ff52afb2-3ce9-4589-ba7f-badb06fa5d51.png)

[Figma](https://www.figma.com/file/Su3VO6yupz4yxe88fv0Uqa/Prebuilt-Components?type=design&node-id=258-26284&t=dKwaWxwLBxFyHpV7-0)

Issue #30, opened 2023-04-19
Access Code View


dawnho:

mikewuu:
```ts
    enableAccessCodeEdit: bool
    enableAccessCodeDelete: bool
```

Can we rename these for brevity?

```
    canEdit: bool
    canDelete: bool
```

seveibar:
`canEdit` and `canDelete` makes sense to me!

razor-x:
I am ok with `canEdit`, `can*` but will offer an alternative of `allowEdit`. 

Thinking of the negation pattern too might help us decide a prefix.  I think the default values of bool props should be false, since that aligns with the undefined value (see also disabled attribute for input: it's not enabled).

`canEdit` / `cannotEdit` vs `allowEdit` / `disallowEdit`?

dawnho:
I borrowed the wording from the MUI component api. Since this is an action (enabling / allowing), a verb is quite appropriate here. Also open to using "can"

dawnho:
So far, most PMS's prefer to disable editing and deleting by default. They were worried about not catching the changes made to the access code and falling out of sync.

PR #114, opened 2023-05-09
Fix access code table layout

### LINKED ISSUE #93: AccessCodes Table - Fix horizontal sizing
Horizontal scroll bar showing when access code table is inside a modal

image

Also there's no more (overflow) button at the end of each row, but I'm not sure if that's meant to be there yet.

image

mikewuu:
Fixes #93 

Was due to `width: 100%` on the header component. Good to remove as any layout should use `flex` to avoid width issues because we're not sure what the parent component would be.

### Note on `AccessCodeTableContent`

After #80 was merged, we lost the ability to preview `AccessCodeTable` in storybook because without a fake it was trying to make real api calls.

As a temporary workaround, I've refactored out `AccessCodeTableContent` which is the UI for `AccessCodeTable`, without the data-binding. I've also updated the story to use `AccessCodeTableContent` for now until the fake client is ready.

### DIFF
--- a/src/lib/ui/AccessCodeTable/AccessCodeTableContent.tsx
+++ b/src/lib/ui/AccessCodeTable/AccessCodeTableContent.tsx
@@ -1,0 +1,61 @@
+import type { AccessCode } from 'seamapi'
+
+import {
+  TableBody,
+  TableCell,
+  TableHeader,
+  TableRow,
+  TableTitle,
+} from 'lib/ui/Table/index.js'
+
+import { AccessCodeKeyIcon } from 'lib/icons/AccessCodeKey.js'
+import { CodeDetails } from 'lib/ui/AccessCodeTable/CodeDetails.js'
+import { ContentHeader } from 'lib/ui/layout/ContentHeader.js'
+import { Caption } from 'lib/ui/typography/Caption.js'
+import { Title } from 'lib/ui/typography/Title.js'
+
+export interface AccessCodeTableContentProps {
+  accessCodes: AccessCode[]
+  onClickBack?: () => void
+}
+
+export function AccessCodeTableContent(
+  props: AccessCodeTableContentProps
+): JSX.Element {
+  const { accessCodes, onClickBack } = props
+
+  return (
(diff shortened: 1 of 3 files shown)


mikewuu:
Oh it was locally! Didn't know the fake was in place.

src/lib/ui/AccessCodeTable/AccessCodeTableContent.tsx
```diff
@@ -0,0 +1,61 @@
...
+              <div>
+                <AccessCodeKeyIcon />
+              </div>
+            </TableCell>
+            <TableCell className='seam-name-cell'>
+              <Title>{code.name}</Title>
+              <CodeDetails accessCode={code} />
+            </TableCell>
+            <TableCell className='seam-action-cell'>
+              {/* <IconButton>
```
razor-x, line 48: Is this attached to an issue?

razor-x (approved):
Note this issue was created to track adding the hook back https://github.com/seamapi/react/issues/108

src/lib/ui/AccessCodeTable/AccessCodeTableContent.tsx
mikewuu: Yep, will be required when either of these are being implemented:

https://github.com/seamapi/react/issues/107
https://github.com/seamapi/react/issues/30

razor-x: Ok, can we remove the comments then? You can paste the code snippet in the issues if they are still relevant at that time.

mikewuu: Damn, merged this a little early. I usually don't leave comments in case they're left dangling for eternity, but it's pretty safe to say this won't be missed -- those 2 features are required.

Do you want me to remove the comments in a separate PR?

razor-x: No worries. As long as they eventually disappear haha

Issue #110, opened 2023-05-09
Device Details - Ability to unlock device


mikewuu:
Currently we're only showing the lock status, and missing the button to unlock/lock:

### Tasks

- [x] Add button
- [ ] Hook up button to mutation

![Image](https://user-images.githubusercontent.com/11449462/236985054-f24811fd-7160-4e07-a5fa-09305691bb4b.png)

[Figma](https://www.figma.com/file/Su3VO6yupz4yxe88fv0Uqa/Prebuilt-Components?type=design&node-id=291-40595&t=dKwaWxwLBxFyHpV7-0)

PR #117, opened 2023-05-10
Buttons & CSS Color Palette

mikewuu:
Partially addresses #110 by adding the `<Button/>` still need to hook it up to a mutation to make the call.

image

### Other Notable Things

- Define all colors in **_colors.scss**
- Define all `$primary` color derivatives as outlined in figma
- Add button styles for all variants / sizes

image

### DIFF
--- a/src/lib/ui/Button.stories.tsx
+++ b/src/lib/ui/Button.stories.tsx
@@ -1,3 +1,4 @@
+import { Box } from '@mui/material'
 import type { Meta, StoryObj } from '@storybook/react'
 
 import { Button } from 'lib/ui/Button.js'
@@ -28,13 +29,7 @@ type Story = StoryObj<typeof Button>
 export const Outline: Story = {
   render: (props) => {
     return (
-      <div
-        style={{
-          display: 'grid',
-          gap: '12px',
-          gridTemplateColumns: 'repeat(3, min-content)',
-        }}
-      >
+      <Box display='grid' gap={3} gridTemplateColumns='repeat(3, min-content)'>
         <div>
           <Button {...props} variant='outline' size='small' />
         </div>
@@ -62,7 +57,7 @@ export const Outline: Story = {
         <div>
           <Button {...props} variant='neutral' size='large' />
         </div>
-      </div>
+      </Box>
     )
   },
 }
--- a/src/lib/ui/Button.stories.tsx
+++ b/src/lib/ui/Button.stories.tsx
@@ -1,0 +1,70 @@
+import type { Meta, StoryObj } from '@storybook/react'
+
+import { Button } from 'lib/ui/Button.js'
+
+const meta: Meta<typeof Button> = {
+  title: 'Library/Button',
+  component: Button,
+  args: {
+    children: 'Label',
+  },
+  tags: ['autodocs'],
+  argTypes: {
+    variant: {
+      table: {
+        disable: true,
+      },
+    },
+    size: {
+      table: {
+        disable: true,
+      },
+    },
+  },
+}
+
+type Story = StoryObj<typeof Button>
+
+export const Outline: Story = {
+  render: (props) => {
+    return (
+      <div
+        style={{
+          display: 'grid',
+          gap: '12px',
+          gridTemplateColumns: 'repeat(3, min-content)',
+        }}
+      >
+        <div>
+          <Button {...props} variant='outline' size='small' />
+        </div>
+        <div>
+          <Button {...props} variant='outline' size='medium' />
+        </div>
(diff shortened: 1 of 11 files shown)


src/lib/ui/Button.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+interface ButtonProps {
+  children?: string
+  variant?: 'solid' | 'outline' | 'neutral'
+  size?: 'small' | 'medium' | 'large'
+  disabled?: boolean
+}
+
+export function Button(props: ButtonProps) {
+  const { variant = 'outline', children, size = 'medium', disabled } = props
```
razor-x, line 11: I feel like for defaults like this we should prefer

```ts
export function Button({ variant = 'outline', children, size = 'medium', disabled }: ButtonProps) {
```

razor-x (approved):

src/styles/_colors.scss
```diff
@@ -0,0 +1,30 @@
+@use 'sass:color';
```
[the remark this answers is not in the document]
mikewuu: > CCS vars seemed like a great solution, but does color.scale($primary, $lightness: 50%); work if $primary is a CSS var?

It does not, as `$primary` is a build-time variable, where `var(--primary)` is runtime.

>  if we wanted to allow any color customization by the consumer, don't know how we would do that now.

Yep, so a couple options:

### 1. Sass

If the consumer is using a bundler that can handle scss, they can override the primary color, and everything else will adjust:

```scss
@use './styles/colors' with (
  $primary: #FC9700
);

@use './styles/main.scss';
```

This was the main reason behind migrating to sass variables, which allows deriving the full set of colors from a single `$primary` color, which was how the [color palette figma](https://www.figma.com/file/Su3VO6yupz4yxe88fv0Uqa/Prebuilt-Components?type=design&node-id=334-57691&t=svZERHeSe6KoTgpK-0) was designed.

### 2. Offer CSS build script

For consumers that don't use sass, maybe we can offer a custom build script?

```
npx seam-react build:css --primary-color #FC9700
```

src/lib/ui/Button.tsx
mikewuu: I usually destructure in a separate line in case we need to forward all the props to a child component. Doing it inline would be a bit more typing:

```tsx
export function Button({ variant = 'outline', children, size = 'medium', disabled, ...otherProps }: ButtonProps) {
   return <SpecialButton  variant={variant} children={children} size={size} disabled={disabled} {...otherProps} />
}
```

src/lib/ui/Button.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+  },
+}
+
+type Story = StoryObj<typeof Button>
+
+export const Outline: Story = {
+  render: (props) => {
+    return (
+      <div
+        style={{
```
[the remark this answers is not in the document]
mikewuu: Sure, updated in https://github.com/seamapi/react/pull/117/commits/d014c213ef8f218e27b9fbce220700ee2341ff46. `sx` is kinda similar to inline styles though, so I went with `Box` props.

src/styles/_colors.scss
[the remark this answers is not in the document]
razor-x: Ok, let's stick with Sass, just trying to think ahead as I know I've hear sever times we will want to offer some minor color customization.

I think the way I've seen this offered in the past is you tell consumers they can build a custom css output by calling our Sass entry point but with a `.sass` file that overrides variables. I'd prefer to ensure we are compatible with the sass cli and not introduce a custom cli tool.

We could also provide a "css var version" where the variable values are all set to css vars and do not use any of the color functions. We could do this later, basically just replace the sass function output with the values.

src/lib/ui/Button.tsx
razor-x: Yea, I have been frustrated by wanting to use these two incompatible patterns at the same time. In this case though we are not forwarding. Also with default values, we can't forward everything since we need to assign the defaults.

My main thought here was having the defaults in the function signature might make for better TS doc magic, but that was just a theory.

mikewuu: >  Also with default values, we can't forward everything since we need to assign the defaults.

Child components would be able to specify their own defaults.

> Yea, I have been frustrated by wanting to use these two incompatible patterns at the same time.

Yep. As long as we're cool with having different patterns when needed I think we can destructure it in function signature in this case.

mikewuu: Updated in https://github.com/seamapi/react/pull/117/commits

Issue #123, opened 2023-05-11
Device Table - Add nav to show Device Detail on row click


mikewuu:
Clicking on a device in the row should show the `<DeviceDetail/>` for the selected device.

![Image](https://user-images.githubusercontent.com/11449462/237747190-62a8c593-daa9-4f67-9105-9d7d84dc6947.png)

Issue #122, opened 2023-05-11
Device Table - show empty placeholder


mikewuu:
Show placeholder message when no devices are found.

![Image](https://user-images.githubusercontent.com/11449462/237746356-d1054102-3909-455e-8214-b222657f1dd0.png)

[Figma](https://www.figma.com/file/Su3VO6yupz4yxe88fv0Uqa/Prebuilt-Components?type=design&node-id=258-26284&t=du5K4KshVcSiPP3C-0)

Issue #121, opened 2023-05-11
Device Table - Ability to search for devices


mikewuu:
Need to add the search input, and implement search:

![Image](https://github.com/seamapi/react/assets/11449462/6366eeec-09be-48dc-9db8-2d9cc849cd2d)

Issue #120, opened 2023-05-11
Device Table - Add device health bar (filter)


mikewuu:
This issue is to add the summary of the device health at the top of the table:

![Image](https://user-images.githubusercontent.com/11449462/237745730-2871e48a-b358-4730-94e9-5b3932249241.png)

[Figma of variants](https://www.figma.com/file/Su3VO6yupz4yxe88fv0Uqa/Prebuilt-Components?type=design&node-id=521-60691&t=du5K4KshVcSiPP3C-0)

- [x] Show status
- [x] Ability to click to filter devices
- [x] Click to reset filters

PR #124, opened 2023-05-11
Device Table

### LINKED ISSUE #27: Devices Table
## Designs

We will not be implementing the modal, as Jurny wants to use their own modal. We'll just be focusing on the innards. I want to get y'all's inputs on how we might tie one view to another

Image
[Link to figma](https://www.figma.com/file/Su3VO6yupz4yxe88fv0Uqa/Out-of-the-box-Components?type=design&node-id=291-40595&t=EyoxpYXT25IB2jin-0)

Note, we're going to implement infinite scroll, as opposed to pagination.

## Properties
Publicly available props for the component:

```
interface DevicesTableProps {
    deviceIds?: Array<string>
    onDeviceRowClick?: (deviceId: string) => void
    sortBy: "name" | "createdAt"
}
```
We'll list out all the devices listed in `deviceIds`. If `deviceIds` is undefined, show all devices.

`onDeviceRowClick` will allow a user to customize what they do when a user clicks on a device row.

## Search

Search should look over the devices' `name` and `location` data.

## Default Sort
We'll sort the devices by `name`.

mikewuu:
Closes #27 

This PR adds the main UI (device table).

image

### Missing features

Anything not included here has been spun off into separate issues:

- https://github.com/seamapi/react/issues/123
- https://github.com/seamapi/react/issues/119
- https://github.com/seamapi/react/issues/120
- https://github.com/seamapi/react/issues/121
- https://github.com/seamapi/react/issues/122
- https://github.com/seamapi/react/issues/106


src/lib/seam/devices/index.ts
```diff
@@ -1 +1,9 @@
+import type { LockDevice } from 'seamapi'
+
 export * from './use-devices.js'
+
+export const getLockModel = ({ properties }: LockDevice) => {
+  return (
+    properties?.august_metadata?.model ?? properties?.schlage_metadata?.model
```
razor-x, line 7: @seveibar Interesting that a consumer of the Seam API would need to implement this exhaustive fallback pattern over all device providers as well as come back and add new ones here. Is there a better interface for this, could there be?

src/lib/seam/devices/index.ts
```diff
@@ -1 +1,9 @@
+import type { LockDevice } from 'seamapi'
+
 export * from './use-devices.js'
+
+export const getLockModel = ({ properties }: LockDevice) => {
```
razor-x, line 5:
```suggestion
export const getDeviceModel = ({ properties }: LockDevice): string | undefined => {
```

src/lib/seam/devices/index.ts
seveibar: glad you mentioned this. the current `*_metadata.model` is a bad attempt at a standardization. Sy has been working to standardize the device db so that we have a representative model. I'm currently of the opinion that model should be a standard object not a standard string e.g. `device.properties.model = { color: "black", year: 2022, series_name: "YRD226" }` or something. But there is a lot of work to be done- it's not an easy problem unfortunately if you want certain assumptions around the model name to hold true (e.g. that two models are the same for purchasing where colors and trim matter vs the same for functionality)

seveibar: i think having logic like this is a great interim solution until we have a proper standardized model (which is a great RFC project)

src/lib/ui/DeviceTable/DeviceTable.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+type Story = StoryObj<typeof DeviceTable>
+
+export const Content: Story = {}
+
+export const InsideModal: Story = {
+  render: InsideModalComponent,
+}
+
+function InsideModalComponent(props: DeviceTableProps) {
+  const [showing, toggleShowing] = useToggle()
```
razor-x, line 133: Maybe math the prop name here (same find/replace for other stories)?
```suggestion
  const [open, toggleOpen] = useToggle()
```

src/lib/ui/DeviceTable/DeviceTable.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { Caption } from 'lib/ui/typography/Caption.js'
+import { Title } from 'lib/ui/typography/Title.js'
+
+export interface DeviceTableProps {
+  devices: LockDevice[]
+  onBack?: () => void
+}
+
+export function DeviceTable({ devices, onBack }: DeviceTableProps) {
+  const numDevices = devices.length
```
razor-x, line 24:
```suggestion
  const deviceCount = devices.length
```

src/lib/ui/DeviceTable/DeviceTable.tsx
```diff
@@ -0,0 +1,63 @@
...
+        {devices.map((device) => (
+          <TableRow key={device.device_id}>
+            <TableCell className='seam-image-cell'>
+              <DeviceImage device={device} />
+            </TableCell>
+            <TableCell className='seam-body-cell'>
+              <Title>{device.properties.name}</Title>
+              <div className='seam-bottom'>
+                <span className='seam-device-model'>
+                  {getLockModel(device) ?? t.unknownLock}
```
razor-x, line 44: Minor point, but can you move this above the return to `const deviceModel =...`

src/lib/seam/devices/index.ts
razor-x, line 1: No code in index files, move to get-device-model or into DeviceModel.tsx

src/lib/ui/DeviceDetails/DeviceModel.tsx
```diff
@@ -0,0 +1,25 @@
+import type { LockDevice } from 'seamapi'
+
+import { getLockModel } from 'lib/index.js'
+
+interface ModelStatusProps {
```
razor-x, line 5:
```suggestion
interface DeviceModelProps {
```

src/lib/ui/DeviceDetails/DeviceModel.tsx
```diff
@@ -0,0 +1,25 @@
...
+  device: LockDevice
+}
+
+export function DeviceModel({ device }: ModelStatusProps): JSX.Element | null {
+  const model = getLockModel(device)
+  if (!model) {
```
razor-x, line 11: The lint rule was unintentionally disabled, I mean to correct that.
```suggestion
  if (model == null) {
```

src/lib/ui/DeviceTable/index.ts
razor-x, line 1: Let's either remove the index files (preferred) or add them to all component modules so there is always a consistent way to import things. Unless there was a reason you were thinking to add it here.

seveibar (approved):

src/lib/ui/device/BatteryStatus.tsx
```diff
@@ -5,32 +5,27 @@ import { BatteryLevelFullIcon } from 'lib/icons/BatteryLevelFull.js'
 import { BatteryLevelHighIcon } from 'lib/icons/BatteryLevelHigh.js'
 import { BatteryLevelLowIcon } from 'lib/icons/BatteryLevelLow.js'
 
-export function BatteryStatus(props: {
-  device: LockDevice
-}): JSX.Element | null {
+export function BatteryStatus(props: { device: LockDevice }) {
```
razor-x, line 8: Prefer `export interface BatteryStatusProps` over inline for the main named component in the module.

src/lib/ui/device/OnlineStatus.tsx
```diff
@@ -4,20 +4,21 @@ import { OnlineStatusDeviceOfflineIcon } from 'lib/icons/OnlineStatusDeviceOffli
 import { OnlineStatusOnlineIcon } from 'lib/icons/OnlineStatusOnline.js'
 
 export function OnlineStatus(props: { device: LockDevice }): JSX.Element {
```
razor-x, line 6: Prefer `export interface OnlineStatusProps` as noted.

src/lib/seam/devices/index.ts
razor-x: Good to know  this is on the radar!

For now, I'm wondering if we could make this function dynamic, e.g., loop through properties, and pick the first non-null `properties?.*_metadata.model`.

razor-x (approved):
Comments are mostly code style and non-blocking. I may merge this early so things can keep moving.

src/lib/seam/devices/index.ts
seveibar: that seems ok- although i'm almost wondering if we could do something like `device.properties.model = { display_name }` where you're not committing to functionality/trim/purchasing.

razor-x: Would this be adding to the hook so it returns a slightly extended type? Or on the API response?

seveibar: sorry on the api response- i think we could propose the model object and push that through relatively quickly.

razor-x: https://github.com/seamapi/react/issues/127

src/lib/seam/devices/index.ts
mikewuu: Updated in https://github.com/seamapi/react/pull/124/commits/990b30a8a4a9e9a1c9418e097898b1d1c1eda2fd

src/lib/ui/DeviceTable/DeviceTable.stories.tsx
mikewuu: Updated in https://github.com/seamapi/react/pull/124/commits/990b30a8a4a9e9a1c9418e097898b1d1c1eda2fd

src/lib/ui/DeviceTable/DeviceTable.tsx
mikewuu: Updated in https://github.com/seamapi/react/pull/124/commits/990b30a8a4a9e9a1c9418e097898b1d1c1eda2fd

src/lib/seam/devices/index.ts
mikewuu: Updated in https://github.com/seamapi/react/pull/124/commits/7e5ee9bdaa2a35fc55e0a797d76609997b669ccc

src/lib/ui/DeviceTable/DeviceTable.tsx
```diff
@@ -21,42 +19,49 @@ export interface DeviceTableProps {
...
-            </TableCell>
-          </TableRow>
+          <DeviceRow device={device} key={device.device_id} />
         ))}
       </TableBody>
     </div>
   )
 }
 
+function DeviceRow(props: { device: LockDevice }) {
```
razor-x, line 41: ❤️ ❤️ ❤️

src/lib/ui/DeviceTable/index.ts
mikewuu: @razor-x oh? wasn't sure if there was a lib-related reason to include index files. Removed in https://github.com/seamapi/react/pull/124/commits/819e3e035dc30998098346dfc8a602852b94a64f

src/lib/ui/device/BatteryStatus.tsx
mikewuu: Updated in https://github.com/seamapi/react/pull/124/commits/72461f87302670be08f6e88bb5968294f9f50b0e

src/lib/ui/device/OnlineStatus.tsx
mikewuu: Updated in https://github.com/seamapi/react/pull/124/commits/72461f87302670be08f6e88bb5968294f9f50b0e

src/lib/ui/device/LockStatus.tsx
```diff
@@ -0,0 +1,42 @@
+import type { LockDevice } from 'seamapi'
+
+import { LockLockedIcon } from 'lib/icons/LockLocked.js'
+import { LockUnlockedIcon } from 'lib/icons/LockUnlocked.js'
+
+export function LockStatus(props: { device: LockDevice }): JSX.Element {
```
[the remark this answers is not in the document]
mikewuu: Updated in https://github.com/seamapi/react/pull/124/commits/72461f87302670be08f6e88bb5968294f9f50b0e

src/lib/ui/DeviceDetails/DeviceModel.tsx
mikewuu: Ok, this is going to take some getting used to, very used to typing `===`. I get the readability of it though, so will try to remember to `== null` instead

razor-x: This is a JS idiosyncrasy that I hate but have accepted as reality. `x === null` is too strict, `x === null || x === undefined` is too verbose, `!x` works for both, but also `!0 === true` so it's too broad. Same thing with `!==` but it's worse, since you end up with `!!x` sometimes. But, `x == null` and `x != null` check for what you need and nothing else, even though it breaks the `===` "rule"

PR #137, opened 2023-05-12
Nested Component Navigation

### LINKED ISSUE #92: Add navigation buttons to handle nested component state
For each component, I want to add the following props:
- `enableBack: bool | (componentName: string, componentProps: object) => void`
  - If set to `true`, we show the back button, if a handler is set, we can call the function whenever the back button is clicked.
- `enableClose: bool | () => void`
  - If set to `true`, we show the close button, if a handler is set, we can call the function whenever the close button is clicked.
- `disableRedirect: bool`
  - will the component redirect to another page when a button is pressed?
  - default to `false`

This will apply to the following components:
- DeviceDetails
- DevicesTable
- AccessCodeDetails
- AccessCodesTable

### LINKED ISSUE #123: Device Table - Add nav to show Device Detail on row click

mikewuu:
Closes #123 
Closes #92 

This PR was contains the original approach to navigation, each component receives an `onBack`, and renders its own child. Unfortunately it ran into Circular Dependency errors in CI, and so #136 separated out the behavior into separate components.

### DIFF
--- a/src/lib/ui/AccessCodeDetails/AccessCodeDetails.tsx
+++ b/src/lib/ui/AccessCodeDetails/AccessCodeDetails.tsx
@@ -1,21 +1,38 @@
 import { DateTime } from 'luxon'
-import type { AccessCode } from 'seamapi'
+import { useState } from 'react'
+import type { AccessCode, CommonDeviceProperties, Device } from 'seamapi'
 
 import AccessCodeDevice from 'lib/ui/AccessCodeDetails/AccessCodeDevice.js'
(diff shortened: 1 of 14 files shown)


mikewuu:
Marking this as **draft** to avoid accidentally merging, but there's no other work to be done here. If we can figure out a way for CI not to throw circular dep errors, this approach might be preferred, and we can merge as is.

@razor-x , thoughts?

PR #138, opened 2023-05-12
Add <NavigationProvider/> to <SeamProvider/>

mikewuu:
- Implicitly include `<NavigationProvider/>`
- Allow `onBack` override at component level

Although this approach offers a better API, it doesn't work well when inside a modal because having `<NavigationProvider/>` at the top-level (included) will re-render all children (including the modal)

https://github.com/seamapi/react/assets/11449462/21278258-f461-49d5-a592-5e0fafd57111


mikewuu:
> just to confirm the onBack is optional, right?

Yep! I think the idea is to allow users to override it if they want.

razor-x:
Closing in favor of https://github.com/seamapi/react/pull/139

PR #153, opened 2023-05-17
enable read-only fake api

seveibar:


vercel.json
```diff
@@ -6,6 +6,12 @@
     "silent": true
   },
   "cleanUrls": true,
+  "rewrites": [
+    {
+      "source": "/api/:fakepath*",
+      "destination": "/api/fakepath"
```
seveibar, line 12: serverless functions are similar to nextjs functions but a bit different. They don't support `[...anypath].ts` style files for example- you have to use a rewrite

razor-x (approved):

PR #154, opened 2023-05-17
Hard-code IDs in storybook

razor-x:
- Remove TODO
- Remove TODO note
- doc: Fix storybook comments
- Move publishableKey and userIdentifierKey to globalTypes

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -23,9 +23,6 @@ const preview: Preview = {
   },
   decorators: [
     (Story) => {
-      // TODO: This should be given by the story, some stories will want
-      // to use different publishable keys, user identifiers etc. to simulate
-      // different scenarios
       return (
         <SeamProvider
           publishableKey={publishableKey}
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -5,6 +5,9 @@ import type { Preview } from '@storybook/react'
 
 import { fakePublishableKey, fakeUserIdentifierKey } from './seed-fake.js'
 
+const isProd = process.env['NODE_ENV'] === 'production'
+const useFake = process.env[']STORYBOOK_SEAM_ENDPOINT'] == null
+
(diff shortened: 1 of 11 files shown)


seveibar (changes requested):
We should be hard coding fake ids into stories. Stories should be deterministic. Multiple development workflows are confusing devs, and add significant complexity.

.storybook/preview.tsx
```diff
@@ -5,13 +5,35 @@ import type { Preview } from '@storybook/react'
 
 import { fakePublishableKey, fakeUserIdentifierKey } from './seed-fake.js'
 
-const publishableKey =
-  process.env['STORYBOOK_SEAM_PUBLISHABLE_KEY'] ?? fakePublishableKey
-
-const userIdentifierKey =
-  process.env['STORYBOOK_SEAM_USER_IDENTIFIER_KEY'] ?? fakeUserIdentifierKey
+const isProd = process.env['NODE_ENV'] === 'production'
+const useFake = process.env[']STORYBOOK_SEAM_ENDPOINT'] == null
```
mikewuu, line 9:
```suggestion
const useFake = process.env['STORYBOOK_SEAM_ENDPOINT'] == null
```

mikewuu, line 9: Is `isProd` & `useFake` mutually exclusive?

mikewuu (approved):

.storybook/preview.tsx
razor-x: No, and I expect these conditionals will be short lived anyway. Is prod is just if it's going to be deployed. We don't have a great way to detect the fake, but the way thing are setup, the fake is always run by storybook and the endpoint port is random, so if you override the endpoint you would not be using the fake, at least not the one provided here.

razor-x: Hmm, this condition is not quite right though, we will need to add the env var back for now.

PR #155, opened 2023-05-17
Always use the fake (unless very explicitly opted out)

seveibar:
Makes the fake an opt-out experience. The production storybook now uses the read-only fake.

The only reason to support the real api at this point for the storybook is because the fake is read only so action attempts can't be polled, webviews can't be finished etc.

We'll have a fake setup that supports modifications soon that will support mutating, so it's a temporary limitation for some of the interactive storybooks.

Making sure everyone develops the same reproducible way is the goal. Some things in the demo don't look right as a result of this- those are bugs that should be fixed. It's more important to have our development environment reflect those bugs then to continue testing against live and having "it works on my system" issues.

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -3,25 +3,24 @@ import '../src/index.scss'
 import { SeamProvider } from '@seamapi/react'
 import type { Preview } from '@storybook/react'
 
-import { fakePublishableKey, fakeUserIdentifierKey } from './seed-fake.js'
-
-const isProd = process.env['NODE_ENV'] === 'production'
-const useFake = process.env['STORYBOOK_SEAM_ENDPOINT'] == null
+// process.env['STORYBOOK_SEAM_ENDPOINT'] == null
+const useFake = true
 
 const preview: Preview = {
   globalTypes: {
     /** @deprecated use "seam_pk1ws2_0000" **/
     publishableKey: {
       description: 'Seam publishable key',
-      defaultValue:
-        process.env['STORYBOOK_SEAM_PUBLISHABLE_KEY'] ?? fakePublishableKey,
+      defaultValue: useFake
+        ? 'seam_pk1ws2_0000'
+        : process.env['STORYBOOK_SEAM_PUBLISHABLE_KEY'],
     },
     /** @deprecated use "seed_client_session_user_2" **/
     userIdentifierKey: {
       description: 'Seam user identifier key',
-      defaultValue:
-        process.env['STORYBOOK_SEAM_USER_IDENTIFIER_KEY'] ??
-        fakeUserIdentifierKey,
+      defaultValue: useFake
+        ? 'seed_client_session_user_2'
+        : process.env['STORYBOOK_SEAM_USER_IDENTIFIER_KEY'],
     },
     /** @deprecated use "device1" **/
     deviceId: {
@@ -37,6 +36,10 @@ const preview: Preview = {
         ? 'access_code1'
         : 'a4d00b36-09e2-467f-a9f3-bb73cea1351c',
     },
+    seamEndpoint: {
+      description: 'Seam Endpoint',
+      defaultValue: useFake ? '/api' : process.env['STORYBOOK_SEAM_ENDPOINT'],
+    },
   },
   parameters: {
     actions: { argTypesRegex: '^on[A-Z].*' },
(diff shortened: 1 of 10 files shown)


.storybook/preview.tsx
```diff
@@ -3,36 +3,44 @@ import '../src/index.scss'
 import { SeamProvider } from '@seamapi/react'
 import type { Preview } from '@storybook/react'
 
-import { fakePublishableKey, fakeUserIdentifierKey } from './seed-fake.js'
-
-const isProd = process.env['NODE_ENV'] === 'production'
-const useFake = process.env['STORYBOOK_SEAM_ENDPOINT'] == null
+const useFake =
+  process.env['DONT_USE_FAKE'] === '1' &&
+  process.env['STORYBOOK_SEAM_ENDPOINT'] == null
```
razor-x, line 8: This won't work since storybook masks all env vars that do not have the storybook prefix. How about?

```suggestion
const useFake =
  ['1','true'].includes(process.env['STORYBOOK_DISABLE_FAKE']?.toLowerCase()) &&
  process.env['STORYBOOK_SEAM_ENDPOINT'] == null
```

.storybook/preview.tsx
```diff
@@ -3,36 +3,44 @@ import '../src/index.scss'
...
     userIdentifierKey: {
       description: 'Seam user identifier key',
-      defaultValue:
-        process.env['STORYBOOK_SEAM_USER_IDENTIFIER_KEY'] ??
-        fakeUserIdentifierKey,
+      defaultValue: useFake
+        ? 'some_user_identifier_key'
+        : process.env['STORYBOOK_SEAM_USER_IDENTIFIER_KEY'],
     },
+    /** @deprecated use "device1" **/
```
razor-x, line 26: Why are these marked `deprecated`? Looks like they are already updated to use the new value.

seveibar: i think stories should use the value directly to avoid abstraction so that people don't understand the scenarios.

razor-x: Ah ok, the comments should suggest that though. So maybe they can say something like `Global IDs will depreciated in favor of values local to each story`

seveibar: agree makes more sense

.storybook/preview.tsx
```diff
@@ -3,36 +3,45 @@ import '../src/index.scss'
 import { SeamProvider } from '@seamapi/react'
 import type { Preview } from '@storybook/react'
 
-import { fakePublishableKey, fakeUserIdentifierKey } from './seed-fake.js'
-
-const isProd = process.env['NODE_ENV'] === 'production'
-const useFake = process.env['STORYBOOK_SEAM_ENDPOINT'] == null
+const useFake =
+  ['1', 'true'].includes(
+    process.env['STORYBOOK_DISABLE_FAKE']?.toLowerCase()
```
razor-x, line 8: Ah, my suggestion broke the types. Here is one fix
```suggestion
  ['1', 'true'].includes(
    process.env['STORYBOOK_DISABLE_FAKE']?.toLowerCase() ?? ''
```

razor-x (approved):

.storybook/preview.tsx
```diff
@@ -3,36 +3,45 @@ import '../src/index.scss'
 import { SeamProvider } from '@seamapi/react'
 import type { Preview } from '@storybook/react'
 
-import { fakePublishableKey, fakeUserIdentifierKey } from './seed-fake.js'
-
-const isProd = process.env['NODE_ENV'] === 'production'
-const useFake = process.env['STORYBOOK_SEAM_ENDPOINT'] == null
+const useFake =
+  ['1', 'true'].includes(
```
seveibar, line 7:
```suggestion
  !['1', 'true'].includes(
```

.storybook/preview.tsx
```diff
@@ -3,36 +3,45 @@ import '../src/index.scss'
...
 import type { Preview } from '@storybook/react'
 
-import { fakePublishableKey, fakeUserIdentifierKey } from './seed-fake.js'
-
-const isProd = process.env['NODE_ENV'] === 'production'
-const useFake = process.env['STORYBOOK_SEAM_ENDPOINT'] == null
+const useFake =
+  !['1', 'true'].includes(
+    process.env['STORYBOOK_DISABLE_FAKE']?.toLowerCase() ?? ''
+  ) && process.env['STORYBOOK_SEAM_ENDPOINT'] == null
```
razor-x, line 9: I think we just need this, not sure why we would have the condition on the endpoint
```suggestion
const useFake =
  !(['1', 'true'].includes(
    process.env['STORYBOOK_DISABLE_FAKE']?.toLowerCase() ?? ''
  )
```

seveibar:
sorry i'm working in suggestions, just trying to figure out the php simultaneously

PR #161, opened 2023-05-19
make it possible to use client session tokens with SeamProvider

seveibar:


mikewuu (approved):

src/lib/SeamProvider.tsx
```diff
@@ -178,11 +182,12 @@ const isSeamProviderPropsWithClientSessionToken = (
     throw new Error('Cannot provide a Seam client along with other options.')
   }
 
-  if ('publishableKey' in props && props.publishableKey != null) {
-    throw new Error(
-      'Cannot provide both a clientSessionToken and a publishableKey .'
-    )
-  }
+  // No reason to ban this behavior, but it's unnecessary.
```
razor-x, line 185: @seveibar The rational to do this is prevent an all to common frustrating scenario where users don't know why something is not working as expected. In this case, a user may wonder why stuff is not working because they think they are only passing one option, or they do not realize which option will 'win'. I recommend we put this check back for better ux.

PR #163, opened 2023-05-19
Access code Table - Add ability to copy code & <Menu/>

### LINKED ISSUE #107: Access Code Table - Add ability to copy code from action menu

mikewuu:
Closes #107 

This PR adds the ability to click-to-copy an Access Code, but it also adds popover menus to the project:

### Click to copy

CleanShot 2023-05-19 at 16 57 28

#### If there's enough space, the default is to show bottom-left

image

#### If there's not enough space on the right, it'll align right instead

image

#### If there isn't enough space on the bottom, it'll show above the anchor

image

#### And it'll also re-calculate, so that it stays in the same position on scroll / window resize

CleanShot 2023-05-19 at 16 56 07


seveibar (approved):

src/lib/ui/Menu/Menu.tsx
```diff
@@ -0,0 +1,179 @@
...
+  verticalOffset = 5,
+  horizontalOffset = 0,
+  edgeOffset = 5,
+  children,
+  button,
+  BackgroundProps,
+}: MenuProps) => {
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null)
+  const [documentEl, setDocumentEl] = useState<null | Element>(null)
+  const [contentEl, setContentEl] = useState<HTMLDivElement | null>(null)
```
razor-x, line 41: @mikewuu Not saying useState is wrong here, but I feel like `useRef` is normally used when dealing with dom element refs (because their lifecycle is outside the react tree lifecycle maybe?)

src/lib/ui/Menu/Menu.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const [top, setTop] = useState(0)
+  const [left, setLeft] = useState(0)
+
+  useEffect(() => {
```
razor-x, line 45: Did we consider using `useLayoutEffect`? Might be appropriate? https://kentcdodds.com/blog/useeffect-vs-uselayouteffect

src/lib/ui/Menu/Menu.tsx
mikewuu: >  lifecycle is outside the react tree lifecycle

Yep. In this case I wanted it to be inside the lifecycle, so we know when to re-calculate.

For side-effects, I'd go with `useRef`.

src/lib/ui/Menu/Menu.tsx
mikewuu: Usually only do `useLayoutEffect` when something seems off due to render timings. Doesn't hurt to switch to it though.

razor-x: Ok, I see you use it below, so maybe this case is correct.

src/lib/ui/Menu/Menu.tsx
```diff
@@ -0,0 +1,179 @@
...
+  useState,
+} from 'react'
+import { createPortal } from 'react-dom'
+
+export interface MenuProps {
+  children: JSX.Element | JSX.Element[]
+  verticalOffset?: number
+  horizontalOffset?: number
+  edgeOffset?: number
+  button: (props: {
```
razor-x, line 17: @mikewuu Should we call this `renderButton`? It confused for a while until I read all the code.

PR #160, opened 2023-05-19
<SupportedDevices /> component

seveibar:

PR #164, opened 2023-05-22
feat: Add `SupportedDevices` component

xplato:
Commandeer of #160 🏴‍☠️☠️ • Adds the `<SupportedDevices />` component for Guesty

---

https://github.com/seamapi/react/assets/87919558/6362a527-ee99-46b5-8fd0-d65da6340830


src/lib/ui/SupportedDevices/SupportedDevices.tsx
```diff
@@ -0,0 +1,17 @@
+import SupportedDeviceRow from './SupportedDeviceRow.js'
+import SupportedDevicesHeader from './SupportedDevicesHeader.js'
+
+export default function SupportedDevices() {
+  return (
+    <table className='seam-supported-devices-table'>
```
xplato, line 6: Using a native `table` element (as well as native `thead`, `th`, `tbody`, `tr`, and `td`, etc.) so that I don't have to do a lot of `!important` rewrites of the current `Table` component. The native browser styling works perfectly for this component. In addition, it avoids a lot of excess imports, etc.

src/lib/ui/SupportedDevices/SupportedDevicesHeader.tsx
```diff
@@ -0,0 +1,14 @@
+export default function SupportedDevicesHeader() {
+  return (
+    <thead>
+      <tr>
+        <th></th>
+        <th>Category</th>
```
xplato, line 6: @seveibar out of curiosity, does Guesty only want locks or will they later include some other devices? Should I add some `props` to allow these columns to be customized?

src/lib/ui/SupportedDevices/SupportedDeviceRow.tsx
```diff
@@ -0,0 +1,21 @@
...
+          style={{ width: 40 }}
+          src='https://www.seam.co/_next/image?url=%2Fimg%2Fdevice-db%2Faugust%2Fsmartlocks%2Faugust_wifi-smart-lock_silver_front.png&w=96&q=75'
+        />
+      </td>
+      <td>Smart Lock</td>
+      <td>Wi-Fi Smart Lock</td>
+      <td>AUG-SL05-M01-S01</td>
+      <td>Wi-Fi</td>
+      <td>
+        <div className="seam-status-pill status-green">
```
xplato, line 15: This isn't in the official design, therefore not an atomic component on its own. This was included in the supported devices table on the marketing site, so I just added a tiny bit of UI and styling to add some color to it. Can easily be removed/extracted out if we like it.

Arc 2023-05-22 at 15 16 47@2x

src/lib/ui/SupportedDevices/SupportedDevicesHeader.tsx
seveibar: We'll add other devices yea, I think "Category" will allow you to switch from e.g. smart locks to thermostats

seveibar: We can probably avoid customization for now i think- but maybe i'll eat my shoe

src/lib/ui/SupportedDevices/SupportedDeviceRow.tsx
seveibar: sgtm

seveibar (approved):

src/styles/_supported-devices.scss
```diff
@@ -0,0 +1,105 @@
...
+        height: 60px;
+        border-bottom: 1px solid colors.$bg-b;
+      }
+
+      td {
+        vertical-align: middle;
+      }
+    }
+
+    // stylelint-disable-next-line no-descending-specificity
```
xplato, line 58: If you move this block above where the linter suggests, it throws more errors saying to move it down. I could probably spend 30-60 minutes rewriting the UI and adding classnames just to make the linter happy, or I could just add this line because I know it's fine and it works.

PR #168, opened 2023-05-23
Fix CSP

seveibar:
The fix from #166 actually broke the whole site, but now I know a bit more about how CSP works yayyy


xplato (approved):
Ooh nice 👌

PR #169, opened 2023-05-23
feat: Add filter and search area to `<SupportedDevices />`

xplato:
Adds an optional filter/search area (visible by default) above the `SupportedDevices` table. Currently, as the video shows, you can filter by brand or active support, or type in a custom search.

The filter by category is currently hidden since the `smartlocks` category is hard-coded in [quick-devicedb-server](https://github.com/seamapi/quick-devicedb-server) (#181)

https://github.com/seamapi/react/assets/87919558/a4aed8d6-57a4-4f1c-b6af-8ba648c03d3c

---

@seveibar still seeing some missing device images in the list unfortunately.


assets/icons/chevron-down.svg
xplato, line 1: I had to manually resize this in Figma because the viewbox was all messed up. Not sure what the issue was (also note the lovely positioning I had to do in the `_supported-devices.scss` file lol).

Side-note: We may want to add a step (if possible) to remove the `id` field from the inner SVG `mask`s because I ran into that issue again where the icon wouldn't display because a duplicate `mask` ID was already in the rendered document. See https://github.com/seamapi/space-frontend/pull/560 (thanks again for finding this @mikewuu!)

src/lib/ui/SupportedDevices/SupportedDevices.tsx
```diff
@@ -1,27 +1,87 @@
...
+
+    if (filterStr.trim() !== '') {
+      queries.push(`text_search=${encodeURIComponent(filterStr.trim())}`)
+    }
+
+    if (filters.supportedOnly) {
+      queries.push('support_level=live')
+    }
+
+    if (filters.category !== null) {
```
xplato, line 40: Not currently used (UI is commented out) because all devices are currently `smartlocks`. Unused, but ready to be.

src/lib/ui/Button.tsx
```diff
@@ -1,11 +1,13 @@
 import classNames from 'classnames'
+import type { MouseEventHandler } from 'react'
 
 interface ButtonProps {
-  children?: string
+  children?: React.ReactNode
```
razor-x, line 5:
```suggestion
```

src/lib/ui/Button.tsx
```diff
@@ -1,11 +1,13 @@
...
   variant?: 'solid' | 'outline' | 'neutral'
   size?: 'small' | 'medium' | 'large'
   disabled?: boolean
-  onClick?: () => void
+  onClick?: MouseEventHandler<HTMLButtonElement>
```
razor-x, line 9: ❤️

src/lib/ui/Button.tsx
```diff
@@ -14,12 +16,17 @@ export function Button({
   size = 'medium',
   disabled,
   onClick,
+  className,
 }: ButtonProps): JSX.Element {
```
razor-x, line 20:
```suggestion
}: PropsWithChildren<ButtonProps>): JSX.Element {
```

src/lib/ui/Button.tsx
```diff
@@ -1,11 +1,13 @@
 import classNames from 'classnames'
+import type { MouseEventHandler } from 'react'
```
razor-x, line 2:
```suggestion
import type { MouseEventHandler, PropsWithChildren } from 'react'
```

src/lib/ui/SupportedDevices/SupportedDevicesFilterArea.tsx
```diff
@@ -0,0 +1,163 @@
+import { capitalize } from '@mui/material'
```
razor-x, line 1: We are not using mui libs in the main components, only in the stories.

src/lib/ui/SupportedDevices/SupportedDevicesFilterArea.tsx
```diff
@@ -0,0 +1,163 @@
...
+  filters,
+  setFilters,
+}: SupportedDevicesFilterAreaProps) {
+  const appliedFilters = Object.keys(filters).filter((key) => {
+    return (
+      filters[key as keyof Filters] !== undefined &&
+      filters[key as keyof Filters] !== null &&
+      filters[key as keyof Filters] !== false
+    )
+  })
```
razor-x, line 33: Does this work?
```suggestion
  const appliedFilters = Object.keys(filters).filter((key: keyof Filters) => {
    return (
      filters[key] != null &&
      filters[key] !== false
    )
  })
```

src/lib/ui/SupportedDevices/SupportedDevicesFilterArea.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { Button } from 'lib/ui/Button.js'
+import Menu from 'lib/ui/Menu/Menu.js'
+import { FilterCategoryMenu } from 'lib/ui/SupportedDevices/FilterCategoryMenu.js'
+import type { DeviceModel, Filters } from 'lib/ui/SupportedDevices/types.js'
+import { TextField } from 'lib/ui/TextField/TextField.js'
+
+interface SupportedDevicesFilterAreaProps {
+  deviceModels: DeviceModel[]
+  filterStr: string
+  setFilterStr: (filterStr: string) => void
```
razor-x, line 15: Prefer to avoid [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation).
```suggestion
  filter: string
  setFilter: (filter: string) => void
```

Or

```
  filterValue: string
  setFilterValue: (filterValue: string) => void
```

src/lib/ui/SupportedDevices/SupportedDevicesFilterArea.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const appliedFilters = Object.keys(filters).filter((key) => {
+    return (
+      filters[key as keyof Filters] !== undefined &&
+      filters[key as keyof Filters] !== null &&
+      filters[key as keyof Filters] !== false
+    )
+  })
+  const appliedFiltersCount = appliedFilters.length
+
+  function getAvailablePropertiesFromDeviceModels(property: keyof DeviceModel) {
```
razor-x, line 36:
```suggestion
  const getAvailablePropertiesFromDeviceModels = (property: keyof DeviceModel) => {
```

src/lib/ui/SupportedDevices/SupportedDevicesFilterArea.tsx
```diff
@@ -0,0 +1,163 @@
...
+    const properties = new Set<string>()
+    deviceModels.forEach((deviceModel) => {
+      properties.add(capitalize(deviceModel[property]))
+    })
+    return Array.from(properties)
+  }
+
+  function resetFilter(filterType: keyof Filters) {
```
razor-x, line 44:
```suggestion
  const resetFilter = (filterType: keyof Filters) => {
```

src/lib/ui/SupportedDevices/FilterCategoryMenu.tsx
```diff
@@ -0,0 +1,63 @@
...
+  onAllOptionSelect,
+  buttonLabel,
+}: FilterCategoryMenuProps) {
+  const usableOptions = addAllOption ? ['All', ...options] : options
+
+  if (Boolean(addAllOption) && onAllOptionSelect == null) {
+    throw new Error(
+      'onAllOptionSelect must be provided if addAllOption is true'
+    )
+  }
```
razor-x, line 28: We would like to avoid throwing errors in components like this. If the props have different modes we could represent this as a union in the prop type.

src/lib/ui/SupportedDevices/SupportedDevicesFilterArea.tsx
```diff
@@ -0,0 +1,163 @@
...
+              Filter
+              {appliedFiltersCount !== 0 ? `s (${appliedFiltersCount})` : ''}
+            </Button>
+          )}
+        >
+          <div
+            className='seam-supported-devices-filter-menu'
+            onClick={(ev) => {
+              ev.stopPropagation()
+            }}
```
razor-x, line 71:
```suggestion
            onClick={(event) => {
              event.stopPropagation()
            }}
```

src/lib/ui/SupportedDevices/SupportedDevicesFilterArea.tsx
```diff
@@ -0,0 +1,163 @@
...
+      <div className='seam-filters-wrap'>
+        <Menu
+          button={({ open }) => (
+            <Button
+              variant='outline'
+              className='seam-filters-button'
+              onClick={open}
+            >
+              Filter
+              {appliedFiltersCount !== 0 ? `s (${appliedFiltersCount})` : ''}
```
razor-x, line 63: Can we follow this pattern? https://github.com/seamapi/react/commit/3eed1df6f4d6ef04e24fbb84876228401bec5120 

See also https://github.com/seamapi/react/pull/174#discussion_r1204966135

src/lib/ui/SupportedDevices/SupportedDevicesFilterArea.tsx
```diff
@@ -0,0 +1,163 @@
...
+                // hard coded—on the backend
+                //
+                // onSelect={(category: string) => {
+                //   setFilters((filters) => ({
+                //     ...filters,
+                //     category,
+                //   }))
+                // }}
+              />
+            </div> */}
```
razor-x, line 91: Can we prefer to file an issue than keep commented code?

src/lib/ui/SupportedDevices/SupportedDevicesFilterArea.tsx
```diff
@@ -0,0 +1,163 @@
...
+                <input
+                  id='supportedOnly'
+                  name='supportedOnly'
+                  type='checkbox'
+                  className='seam-filter-checkbox'
+                  checked={filters.supportedOnly}
+                  onChange={(ev) => {
+                    setFilters((filters) => ({
+                      ...filters,
+                      supportedOnly: ev.target.checked,
```
razor-x, line 125:
```suggestion
                  onChange={(event) => {
                    setFilters((filters) => ({
                      ...filters,
                      supportedOnly: event.target.checked,
```

src/lib/ui/SupportedDevices/SupportedDevices.tsx
```diff
@@ -43,23 +103,48 @@ export function SupportedDevices() {
...
+          </table>
+        )}
       </div>
     </>
   )
 }
+
+SupportedDevices.defaultProps = {
+  showFilterArea: true,
+}
```
razor-x, line 150: I think we can prefer to use 

```
export function SupportedDevices({ showFilterArea = true }: SupportedDevicesProps) {
```

razor-x (commented):
There are still some file conflicts and CI errors, so I can review the conflicting files when those are resolved. Left some comments on other stuff to get started.

razor-x:
@mikewuu and @xplato I'm worried we now have two filter / search implementations now that this is merged https://github.com/seamapi/react/pull/180. Should we consolidate them somehow?

xplato:
> @mikewuu and @xplato I'm worried we now have two filter / search implementations now that this is merged https://github.com/seamapi/react/pull/180. Should we consolidate them somehow?

Yep, just saw this. I'll import Mike's `SearchTextField`

src/lib/ui/SupportedDevices/SupportedDevicesFilterArea.tsx
xplato: Better idea

src/lib/ui/SupportedDevices/SupportedDevicesFilterArea.tsx
xplato: Not easily, no. I'll assign the casted value to a variable to reduce repetition, though.

seveibar (approved):
lgtm, RE: importing mui, yes we should avoid it- it can probably removed entirely from this project soon because I think we'll have a Dialog which was the main reason for importing it

src/lib/ui/Button.tsx
xplato: Oh nice, didn't know this

src/lib/ui/SupportedDevices/SupportedDevices.tsx
xplato: #181

src/lib/ui/SupportedDevices/SupportedDevices.tsx
```diff
@@ -1,27 +1,89 @@
...
+import type { DeviceModel, Filters } from 'lib/ui/SupportedDevices/types.js'
 
 import { SupportedDeviceRow } from './SupportedDeviceRow.js'
 import { SupportedDevicesHeader } from './SupportedDevicesHeader.js'
 
-export function SupportedDevices() {
+const BASE_URL = 'https://devicedb.seam.co/api/device_models/list'
+
+export interface SupportedDevicesProps {
+  // If true, show the filter area and search bar
```
razor-x, line 16:
```suggestion
```

src/lib/ui/SupportedDevices/SupportedDevices.tsx
```diff
@@ -1,27 +1,89 @@
...
+  showFilterArea?: boolean
+}
+
+export function SupportedDevices({
+  showFilterArea = true,
```
razor-x, line 21:
```suggestion
  cannotFilter,
```

src/lib/ui/SupportedDevices/SupportedDevices.tsx
```diff
@@ -43,22 +105,43 @@ export function SupportedDevices() {
...
+              {deviceModels.length === 0 && (
+                <tr className='seam-supported-devices-table-message-row'>
+                  <td colSpan={6}>
+                    <div className='seam-supported-devices-table-message'>
+                      {filterValue.length === 0 ? (
+                        <p>No device models found.</p>
+                      ) : (
+                        <>
+                          <p>No device models matched your search.</p>
+                          <Link
```
razor-x, line 129: Can we replace this with `a` or otherwise?

xplato: Replaced with small `Button` from this project.

razor-x (approved):

src/lib/ui/SupportedDevices/SupportedDevicesTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+      data?.data?.device_models !== undefined &&
+      allDeviceModels.length === 0
+    ) {
+      setAllDeviceModels(data.data.device_models)
+    }
+  }, [data, allDeviceModels.length])
+
+  const deviceModels = data?.data?.device_models ?? []
+
+  const showFilterArea = cannotFilter == null ? true : !cannotFilter
```
xplato, line 72: Personally, I find this much more confusing than simply `!Boolean(cannotFilter)` or just `!cannotFilter`. We might want to consider disabling the `@typescript-eslint/strict-boolean-expressions` linting rule because of how often it causes hiccups

xplato: Okay. Updated the props to make it default to `false`

xplato:
@razor-x re-requested review for good measure—all files were renamed so progress on the review side was reset. My latest commits have good names for going through it that way

src/lib/ui/SupportedDeviceTable/SupportedDeviceTable.tsx
```diff
@@ -0,0 +1,14 @@
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+
+import type { SupportedDeviceContentProps } from 'lib/ui/SupportedDeviceTable/SupportedDeviceContent.js'
+import { SupportedDeviceContent } from 'lib/ui/SupportedDeviceTable/SupportedDeviceContent.js'
+
+const client = new QueryClient()
+
```
razor-x, line 7: We are exporting props for users for our public components like how mui does it. It's nice to have as a dev :)

```suggestion
export type SupportedDeviceTableProps = SupportedDeviceContentProps

```

razor-x (approved):
Thanks, just the one minor comment after the rename 🚀

src/lib/ui/SupportedDeviceTable/SupportedDeviceTable.tsx
xplato: Oops, forgot this one. Thanks! 👍

PR #184, opened 2023-05-27
docs: Proxy unpkg

razor-x:
- Allows us to provide a stable URL for assets independent of the backing CDN.
- Font's still need to come from Google Fonts as re-hosting that is non-trivial (also not versioned, so no need to rehost).
- Once this is on main, can update the internal URL and fix the version replacement at build time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,10 +69,7 @@ import '@seamapi/react/index.css'
 or place the following in the `<head>` tag:
 
 ```html
-<link
-  rel="stylesheet"
-  href="https://react.seam.co/v/^1.0.0/index.min.css"
-/>
+<link rel="stylesheet" href="https://react.seam.co/v/^1.0.0/index.min.css" />
 ```
 
 > You can replace the version string above with the exact version of this package used by your application.
--- a/README.md
+++ b/README.md
@@ -58,8 +58,9 @@ Components are styled with plain Cascading Style Sheets (CSS).
 All styles are prefixed with `seam-` to avoid name collisions.
 
 By default, the `SeamProvider` will inject a link tag into the document head to load the CSS.
-This may be disabled with the `disableCssInjection` prop.
-If you prefer to manually load the CSS, either import it using a supported bundler with
+If you prefer to manually load the CSS,
+this behavior may be disabled with the `disableCssInjection` prop.
+Then, either import the CSS using a supported bundler with
 
 ```ts
 import '@seamapi/react/index.css'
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -50,12 +50,40 @@ export const App = () => {
...
+By default, the `SeamProvider` will inject a link tag into the document head to load the CSS.
+If you prefer to manually load the CSS,
+this behavior may be disabled with the `disableCssInjection` prop.
+Then, either import the CSS using a supported bundler with
+
+```ts
+import '@seamapi/react/index.css'
+```
+
+or place the following in the `<head>` tag:
```
seveibar, line 69: ❤️

vercel.json
```diff
@@ -7,6 +7,10 @@
   },
   "cleanUrls": true,
   "rewrites": [
+    {
+      "source": "/v/:version/:asset*",
+      "destination": "https://www.unpkg.com/@seamapi/react@:version/:asset"
+    },
```
seveibar, line 13: can't remember if this works

```suggestion
    {
      "source": "/v/:version/:asset*",
      "destination": "https://www.unpkg.com/@seamapi/react@:version/:asset",
      "headers": [
          {
            "key": "Cache-Control",
            "value": "s-maxage= 604800"
          }
        ]
    },
```

seveibar: i think it should https://vercel.com/guides/how-to-configure-the-cache-control-response-header-in-vercel-projects#using-the-vercel.json-file

seveibar (approved):

vercel.json
seveibar: hmm, unclear, but worth noting that it'll make things faster if we can use the Cache-Control header

razor-x: We should! Good call! There are a standard set of "Asset CDN" headers I like to ensure are set and forgot to consider them given we are proxing. I think in this case we can respect unpkg + Vercel for now as they seem to be adding this for us without additional config:

image

image

razor-x: Actually, I am more concerned about everything else (that is not an immutable asset), so made an issue to review headers later: https://github.com/seamapi/react/issues/186

seveibar: hmm, Vercel couldn't be adding Cache-Control automatically

seveibar: oh i think i see nvm

seveibar: So unpkg adds cache-control, Vercel sees that and it's the same as us providing it. Gotcha

PR #185, opened 2023-05-27
Minor prop and name fixes

razor-x:
- fix: Use class over style for menu visibility
- fix: Rename seam-supported-devices-table class to match component
- fix: Use img width over style width
- refactor: Prefer PropsWithChildren
- refactor: Prefer && over ternary
- fix: Ensure optional bool props always default to false
- fix: Do not use defaultProps
- refactor: Simplify alertProps

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,19 @@ Then, load it from Google Fonts by placing the following in the `<head>` tag:
 
 ### Content Security Policy (CSP)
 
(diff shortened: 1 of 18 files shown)


razor-x:
@mikewuu Note that there is some more refactor I'd like to do on the supported devices table (like export the hook), but I want to get this in first to limit the changeset and get the fix in for the loading behavior.

mikewuu (approved):

PR #187, opened 2023-05-27
Use correct CSS version when injecting link tag

razor-x:
- refactor: Prefer disabled over enabled
- refactor: Return early from effect
- docs: Import CSS in example
- feat: Add useUnminifiedCss prop
- fix: Use correct version for stylesheet

### DIFF
--- a/examples/basic/src/App.tsx
+++ b/examples/basic/src/App.tsx
@@ -6,6 +6,7 @@ export const App = (): JSX.Element => {
       endpoint={import.meta.env.SEAM_ENDPOINT}
       publishableKey={import.meta.env.SEAM_PUBLISHABLE_KEY}
       userIdentifierKey={import.meta.env.SEAM_USER_IDENTIFIER_KEY}
+      disableCssInjection
     >
       <main>
         <h1>Seam Components</h1>
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "lint": "eslint --ignore-path .gitignore --ext .js,.cjs,.mjs,.ts,.tsx .",
     "prelint": "prettier --check --ignore-path .gitignore .",
     "postlint": "stylelint '**/*.scss'",
+    "prepack": "echo \"export default '$(jq -r .version < package.json)'\" > lib/version.js",
     "postversion": "git push --follow-tags",
     "storybook": "storybook dev --port 6006",
     "storybook:docs": "storybook dev --docs --port 6007",
--- a/src/lib/seam/use-seam-styles.ts
+++ b/src/lib/seam/use-seam-styles.ts
@@ -6,10 +6,12 @@ const version = '1.1.0'
 
 export const useSeamStyles = ({
   disabled = false,
+  unminified = false,
 }: {
+  unminified?: boolean
   disabled?: boolean
 }) => {
-  const cssUrl = `${origin}/v/${version}/index.css`
+  const cssUrl = `${origin}/v/${version}/index${unminified ? '' : '.min'}.css`
 
   useEffect(() => {
     if (version === null) return
--- a/src/lib/seam/use-seam-styles.ts
+++ b/src/lib/seam/use-seam-styles.ts
@@ -1,21 +1,28 @@
 import { useEffect } from 'react'
 
-// TODO figure out the version automatically
-const cssUrl = 'https://www.unpkg.com/@seamapi/react@1.1.0/index.min.css'
+const origin = 'https://react.seam.co'
 
-export const useSeamStyles = ({ disabled = false }: { disabled?: boolean }) => {
+const version = '1.1.0'
+
+export const useSeamStyles = ({
+  disabled = false,
+}: {
+  disabled?: boolean
+}) => {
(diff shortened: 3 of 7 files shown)


examples/basic/src/App.tsx
```diff
@@ -6,6 +6,7 @@ export const App = (): JSX.Element => {
       endpoint={import.meta.env.SEAM_ENDPOINT}
       publishableKey={import.meta.env.SEAM_PUBLISHABLE_KEY}
       userIdentifierKey={import.meta.env.SEAM_USER_IDENTIFIER_KEY}
+      disableCssInjection
```
razor-x, line 9: We have to put this in the example since when running locally the example should use the latest styles. We can keep the font injection.

package.json
```diff
@@ -65,6 +65,7 @@
     "lint": "eslint --ignore-path .gitignore --ext .js,.cjs,.mjs,.ts,.tsx .",
     "prelint": "prettier --check --ignore-path .gitignore .",
     "postlint": "stylelint '**/*.scss'",
+    "prepack": "echo \"export default '$(jq -r .version < package.json)'\" > lib/version.js",
```
razor-x, line 68: Normally I would like to use a platform independent script (can always be done later), but this will ONLY run on CI so it should be safe to do this for now.

src/lib/seam/use-seam-styles.ts
```diff
@@ -1,21 +1,31 @@
...
+}: {
+  unminified?: boolean
+  disabled?: boolean
+}) => {
+  const ext = `${unminified ? '' : 'min.'}css`
+  const cssUrl = `${origin}/v/${version ?? ''}/index.${ext}`
 
-export const useSeamStyles = ({ enabled }: { enabled: boolean }) => {
   useEffect(() => {
+    if (version === null) return
```
razor-x, line 18: In development the version will always be null.

razor-x:
UI test for the supported device table is flaky, seems it get stuck loading sometimes.

seveibar (approved):
fwiw i was testing this w/ `yalc` to make sure the imports go through properly, it's a bit hard to test, maybe you know a better way

src/lib/seam/use-seam-styles.ts
```diff
@@ -1,21 +1,31 @@
 import { useEffect } from 'react'
 
-// TODO figure out the version automatically
-const cssUrl = 'https://www.unpkg.com/@seamapi/react@1.1.0/index.min.css'
+import version from 'lib/version.js'
+
+const origin = 'https://react.seam.co'
```
mikewuu, line 5: If we always use the css file in the release / bundle, wouldn't it always have the correct version?

mikewuu (approved):

src/lib/seam/use-seam-styles.ts
razor-x: No because we are injecting a URL that contains the version number which must match the version of the package that is being used by the user. I'm not sure if that explanation is more clear, but it's correct.

razor-x:
> fwiw i was testing this w/ yalc to make sure the imports go through properly, it's a bit hard to test, maybe you know a better way

@seveibar It's a little difficult to test things that only happen in the built package (very few things that hit that edge case). It's possible though: the `install` CI action actually ensures the package will be installable and the all it's modules will resolve.

This case it pretty special though. We could add an example app that ONLY uses the deployed version (would not "sync" with local dev). That would give a way to test it (I basically did that locally to verify this will work).

razor-x:
@seveibar https://github.com/seamapi/react/issues/189

src/lib/seam/use-seam-styles.ts
mikewuu: > No because we are injecting a URL that contains the version number 

Yep, was wondering what the benefits of injecting via URL vs. just including what is bundled with the package are?

razor-x: Ah, there is a pretty long internal discussion about this. The short version is that having it inject the fonts and the styles simplifies adoption and removes all the extra steps you need to use this. 

It's still possible to opt out of the HTML injection and do it the other way for users who prefer that.

mikewuu: Was actually thinking of dynamically importing the css via relative paths, but of course that wouldn't work because this is runtime.

What if we always include the CSS, but just change the class name? Won't have to deal with version numbers, and no dynamic loading necessary?

```tsx
const class = disableStyles ? '' : 'seam-components'

return (<div className={class}>...</div>)
```

razor-x: > What if we always include the CSS

I'm not sure I understand this. With this update we are always including the css unless the user opts out. I'd prefer not to inject the via an inline style tag since that bloats the js bundle size and cannot take advantage of caching the same way (among other things).

mikewuu: >  I'd prefer not to inject the via an inline style tag

Yep, I mean we'd be able to avoid that. Here's a PR:

https://github.com/seamapi/react/pull/191

PR #182, opened 2023-05-26
feat: Automatic style loading

seveibar:
Simplifies API examples because you don't need to import CSS or a font

https://hello-seam.slack.com/archives/C051ETBPRTQ/p1685131057245149

## Future Work

- [ ] Load the relevant version of styles properly from CDN
- [ ] Add content hash to link tag

PR #191, opened 2023-05-28
fix: avoid style injection

mikewuu:
An alternative approach to the one taken in https://github.com/seamapi/react/pull/182, which offers automatic style loading without versioning, or style injections:

- Includes styles by default.
- Always import `index.css` in `<SeamProvider/>`.
- Allows the user to opt-out by specifying `disableStyles`.

CleanShot 2023-05-28 at 17 02 43


razor-x:
I'd like to avoid this approach as it requires a bundler that understands how to bundle css. As-is, the lib compiles with TSC to a standard es module which is very compatible. I am looking at turning these into web components and I think this change will further complicate that.

mikewuu:
Fair enough. Still wrapping my head around the fact that this is a lib, and we don't have a bundler.

razor-x:
No worries. Even if we used a bundler that could import css, then you have to ask: what would it replace the import with at bundle time? It would have to either be inline styles or a link to a cdn hosted stylesheet 😁

mikewuu:
Don't bundlers just combine all css imports into a single index.css file?

To your point, guess we can't assume that.

razor-x:
> Don't bundlers just combine all css imports into a single index.css file?

That is exactly what Sass cli is doing right now.

PR #197, opened 2023-05-30
Add hooks entrypoint

razor-x:
This moves some things around to better separate internal components from the components in the public API. Adds a hooks entrypoint to expose needed types and functions for integrating without using any react components. This helps with setting up the web component entrypoint. 

- refactor: Rename capitalize module to strings
- refactor: Move ui specific methods to ui
- refactor: Move date sort to dates module
- refactor: Move SeamProvider into seam
- refactor: Move public components to src/lib/seam/components
- feat: Export hooks separately


xplato (approved):
LGTM!

PR #201, opened 2023-06-01
Load device models via SDK

razor-x:
- Supported device table now uses the SDK to load the device models (hook included).
- The fake serves this endpoint to the storybook now which should help with the loading state issues. 
- Handles the enum schema updates.
- Refactored the components a bit to use the hook and flatten the nested logic a bit.
- Fixed a few minor bugs.  

Commits: 
- docs: Update fake-seam-connect
- refactor: Prefer arrow function
- feat: Add useDeviceModels
- refactor: Name FilterCategoryMenuProps interfaces
- refactor: Move text to t for FilterCategoryMenu
- refactor: Move SupportedDeviceFilterArea text to t
- fix: Show number of applies filters when 1 is active
- feat: Use useDeviceModels in SupportedDeviceTable
- Update fake-seam-connect
- refactor: Flatten SupportedDeviceTable render logic
- docs: Add story for SupportedDeviceTable cannotFilter prop
- refactor: Use t for SupportedDeviceHeader
- refactor: Import with lib/ path
- Format connection_type and support_level
- refactor: Pass all label to FilterCategoryMenu
- feat: Update CSP
- docs: Add note on seed data to README


src/lib/seam/components/SupportedDeviceTable/SupportedDeviceContent.tsx
```diff
@@ -1,138 +1,111 @@
...
-          filters={filters}
-          setFilters={setFilters}
-        />
-      )}
+    <table className='seam-supported-device-table-content'>
+      <SupportedDeviceHeader />
+      <tbody>
+        {deviceModels.map((deviceModel) => (
+          <SupportedDeviceRow
+            key={deviceModel.manufacturer_model_id}
```
razor-x, line 60: Is `manufacturer_model_id` unique? Will we eventually have a proper unique id to use here?

src/lib/seam/components/SupportedDeviceTable/SupportedDeviceRow.tsx
```diff
@@ -18,20 +19,30 @@ export function SupportedDeviceRow({
...
   )
 }
 
-const supportLevelColors = {
-  Live: 'green',
-  Beta: 'blue',
-  Unsupported: 'orange',
+const supportLevelColors: Record<
+  DeviceModel['support_level'],
+  'green' | 'blue' | 'orange'
```
razor-x, line 36: Could be `string` but this shows we don't have type protection when mapping to dynamic class names.

src/lib/seam/components/SupportedDeviceTable/SupportedDeviceContent.tsx
xplato: For now I'd probably do `key={`${index}:${deviceModel.manufacturer_model_id}`}` or something to that effect because IIRC I ran into a duplicate key issue. We could have the backend generate a UUID when it parses the CSV file as a more permanent solution.

xplato (approved):
Nice, lgtm!

PR #207, opened 2023-06-02
feat: Add className to components

razor-x:


README.md
```diff
@@ -74,6 +74,17 @@ or place the following in the `<head>` tag:
 <link rel="stylesheet" href="https://react.seam.co/v/1.0.0/index.min.css" />
 ```
 
+#### Customizing Styles
+
+> The `seam-` prefixed CSS class names are not part of the stable API and may change unexpectedly.
```
xplato, line 79: Oh I didn't know this. I thought it was just a practice in this lib. Good to know

xplato (approved):
Awesome; ❤️ to see it

README.md
razor-x: Just to add some context, we prefix all the class names in this lib with `seam-` so external global styles do not affect our components. This line just means we are not making any promises (yet) on how to customize the component styles, so if users attempt to apply styles to our `seam-` classes, that is not technically supported by this lib.

Issue #211, opened 2023-06-05
Mark files with side effects 


razor-x:
https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

PR #212, opened 2023-06-05
Mark files with side effects

razor-x:
Needed for tools that use Webpack (like Angular). Closes https://github.com/seamapi/react/issues/211

- refactor: Build CSS to dist
- fix: Mark elements and css as having side effects
- refactor: Simplify prepack command


razor-x:
Tested this with Angular and made a few tweaks to get it to work.

mikewuu (approved):

PR #221, opened 2023-06-06
feat: Add onAccessCodeClick and preventDefaultOnAccessCodeClick prop to AccessCodeTable

razor-x:


src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
```diff
@@ -85,17 +87,20 @@ export function AccessCodeTable({
         />
       </TableHeader>
       <TableBody>
-        <Body accessCodes={filteredCodes} selectAccessCode={selectAccessCode} />
+        <Body
+          accessCodes={filteredCodes}
+          onAccessCodeClick={onAccessCodeClick ?? setSelectedAccessCodeId}
```
xplato, line 92: I'm not entirely in the loop on this particular feature, but I'm wondering if it might be more consumer-friendly to pair up the event handler here with the default selection behind a boolean prop; for instance, adding a new `shouldSelectAccessCodeRowOnClick` or something to that effect so that just by passing an event handler it doesn't take away the default UX, etc.

I can think of a use-case like e.g. analytics where they want to know a row was clicked but still allow the details to be shown. Just thinking out loud here, pretty minor either way I think.

Ofc, we could always wait until this feature is requested / throw this in an issue for later. Up to you!

xplato (approved):

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
razor-x: This is a good point. Basically we should have a way to `preventDefault`.

razor-x: How about `accessCodeClickPreventDefault`? Then the default behavior needs to be explicitly disabled.

xplato: Yep, that sounds good to me!

PR #230, opened 2023-06-13
Generalize beyond LockDevice

razor-x:
Fixes some minor issues and gets things ready to handle Thermostats.

Items to note in bold (note, ref links work and are accurate diffs but are now out of sync with this PR because I had to rebase to fix the CI triggers)

- 772d713 refactor: Reorder DeviceDetails device type conditional
- a7d47e3 refactor: Prefer Content consistently for private components
- 49ddd54 refactor: Expand each query check
- 462bdd1 refactor: Generalize where Device should be used over LockDevice
- **3f64498 fix: Use battery.status over battery_level**
- fc1f2b3 refactor: Move LockDeviceDetails to module
- d0f03fc refactor: Move isLockDevice check to DeviceRow
- ed8fa1c refactor: Ensure only lock devices appear in DeviceTable
- **ab2d790 fix: Properly detect lock devices**
- 8db81f1 refactor: Prefer to use CommonDevice
- fc04ed3 Run format
- 9215fe8 Run npm update
- 92c2c00 Lower battery_level in seed
- **55cbfb3 feat: Update seamapi**
- **6d9fc5a feat: Use device.properties.model.display_name**
- **2a7bd17 fix: Prefer supported_code_lengths over access_code_length**
- **9a9f141 feat: Display access code length range**


src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
```diff
@@ -35,148 +22,15 @@ export function DeviceDetails({
...
-  if (!isLockDevice(device)) {
-    return null
-  }
-
-  return (
-    <LockDeviceDetails className={className} device={device} onBack={onBack} />
-  )
-}
-
-function LockDeviceDetails(props: {
```
razor-x, line 47: This was moved unchanged to to `src/lib/seam/components/DeviceDetails/LockDeviceDetails.tsx`

src/lib/seam/components/DeviceDetails/LockDeviceDetails.tsx
razor-x, line 1: This was moved unchanged from  `src/lib/seam/components/DeviceDetails/DeviceDetails.tsx`

src/lib/seam/components/DeviceDetails/LockDeviceDetails.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const { accessCodes } = useAccessCodes({
+    device_id: device.device_id,
+  })
+
+  const lockStatus = device.properties.locked ? t.locked : t.unlocked
+  const toggleLockLabel = device.properties.locked ? t.unlock : t.lock
+
+  const accessCodeCount = accessCodes?.length
+
+  const accessCodeLength = device.properties?.supported_code_lengths?.[0]
```
razor-x, line 34: I think this is better logic than 

```
const accessCodeLength = device.properties?.schlage_metadata?.access_code_length
```

codetheweb: yes, should always use the top-level property if available

I think `device.properties?.schlage_metadata?.access_code_length` is an artifact from before we had a standard top-level property

razor-x: Actually, I thought a bit more and we can get min / max do to `a range :)

src/lib/seam/components/DeviceDetails/LockDeviceDetails.tsx
```diff
@@ -0,0 +1,158 @@
...
+  lock: 'Lock',
+  locked: 'Locked',
+  unlocked: 'Unlocked',
+  accessCodes: 'access codes',
+  codeLength: 'Code length',
+  digits: 'digits',
+  lockStatus: 'Lock status',
+  status: 'Status',
+  power: 'Power',
+}
```
seveibar, line 158: :penguin: 

ha pseudo-i18n, fwiw I don't love the enum-oriented pattern because I think it makes contribution harder (project search often begins with a string seen in the UI and if enum strings are present it's a 2+ step search, especially if i18n usage isn't obvious), pretty harmless in this instance. This project is likely to have i18n I think, so I'm not super opposed, but I do want to be cautious of a ton of forward planning around i18n (I regret implementing i18n on https://udt.dev, because it created a slew of issues and made contribution pretty difficult, even though we saw a lot of adoption as a result I don't think it was worth the velocity).

I think if we could implement it such that english-first search is still effective e.g.

```
t("digits")
t("Device")
t("Status")
```

then always keep those inline. This is a bit easier because there's no redefining the english translation. The con is you have to repeat english for long strings, but this can be mitigated by supported a shorthand mapping e.g. `t("terms-of-service")` can still map to a large string

seveibar (approved):

src/lib/seam/components/DeviceDetails/LockDeviceDetails.tsx
razor-x: Yea, another issue with the object is that unused keys are not linted against. This pseudo i18n has three nice features: first, it keeps the text out of the JSX which often makes the code harder to read / creates annoying diffs / makes typos harder to spot, etc; and second, the text is not that far away from where it is used since it stays in the same file alongside the component (which partially mitigates your primary gripe); third, it isolates the localized text so it may be factored out later.

seveibar: ok i'm down to try it!

PR #235, opened 2023-06-27
fix: Do not export useDeviceModels hook

razor-x:
This hooks is unstable and should not have been exported.

### DIFF
--- a/src/lib/seam/device-models/use-device-models.ts
+++ b/src/lib/seam/device-models/use-device-models.ts
@@ -13,7 +13,10 @@ export type UseDeviceModelsParams = DeviceModelsListRequest
 export type UseDeviceModelsData = DeviceModel[]
 
 export function useDeviceModels(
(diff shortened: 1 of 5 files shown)


src/lib/seam/device-models/use-device-models.ts
```diff
@@ -13,7 +13,10 @@ export type UseDeviceModelsParams = DeviceModelsListRequest
 export type UseDeviceModelsData = DeviceModel[]
 
 export function useDeviceModels(
-  params?: UseDeviceModelsParams
+  params?: Omit<
+    UseDeviceModelsParams,
+    'acknowledge_intentional_use_of_internal_api'
```
xplato, line 18: Haha, nice. Getting `dangerouslySetInnerHTML={{ __html: data }}` vibes

xplato (approved):

Issue #238, opened 2023-06-27
`<DeviceDetails />` - Add `disableLockUnlock` prop


dawnho:
This will hide the "Lock" or "Unlock" button for a device.

PR #242, opened 2023-06-27
feat: Add disableFilter and deprecate cannotFilter

razor-x:
Prop convention is moving to `disable/enable` prefixes, see e.g., https://github.com/seamapi/react/issues/238


xplato (approved):

src/lib/seam/components/SupportedDeviceTable/SupportedDeviceTable.tsx
```diff
@@ -26,8 +26,7 @@ export function SupportedDeviceTable({
     brand: null,
   })
 
-  const hideFilter =
-    cannotFilter == null ? disableFilter ?? false : cannotFilter ?? false
+  const hideFilter = cannotFilter ?? disableFilter
```
xplato, line 29: ❤️

PR #246, opened 2023-06-28
Add disableLockUnlock props

razor-x:


xplato:
🏴‍☠️ Taking over in favor of #250

PR #250, opened 2023-06-28
feat: Add `disableLockUnlock` property

### LINKED ISSUE #237: `<AccessCodeDetails />` - Add `disableLockUnlock` prop
- [ ] Add `disableLockUnlock` prop, and Hide Device Lock / Unlock button when `disableLockUnlock` is true

xplato:
Closes #237 && #238

### DIFF
--- a/src/lib/seam/components/AccessCodeDetails/AccessCodeDetails.tsx
+++ b/src/lib/seam/components/AccessCodeDetails/AccessCodeDetails.tsx
@@ -12,12 +12,14 @@ import { useIsDateInPast } from 'lib/ui/use-is-date-in-past.js'
 export interface AccessCodeDetailsProps {
   accessCodeId: string
   onBack?: () => void
+  disableLockUnlock?: boolean
   className?: string
 }
 
 export function AccessCodeDetails({
   accessCodeId,
   onBack,
+  disableLockUnlock = false,
   className,
 }: AccessCodeDetailsProps): JSX.Element | null {
   const { accessCode } = useAccessCode(accessCodeId)
@@ -55,6 +57,7 @@ export function AccessCodeDetails({
         </div>
         <AccessCodeDevice
           deviceId={accessCode.device_id}
+          disableLockUnlock={disableLockUnlock}
           onSelectDevice={selectDevice}
         />
       </div>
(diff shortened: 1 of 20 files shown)


src/lib/seam/components/AccessCodeDetails/AccessCodeDetails.tsx
```diff
@@ -12,12 +12,14 @@ import { useIsDateInPast } from 'lib/ui/use-is-date-in-past.js'
...
   accessCodeId: string
   onBack?: () => void
+  disableLockUnlock?: boolean
   className?: string
 }
 
 export function AccessCodeDetails({
   accessCodeId,
   onBack,
+  disableLockUnlock = false,
```
xplato, line 22: Not sure if these tend to be used individually—if not, we can remove this default value

razor-x: Default is nice at the top level API for props where undefined is not a meaningful value. This means internal components will always be passed a definite value and since it's a required prop on the internal components the type checker will ensure we don't forget to pass drill it down

xplato: I'm keeping the default under the `DeviceDetails` component for now but let me know if it should be added elsewhere.

razor-x: These are top level because they are exported and documented as the public component api https://github.com/seamapi/react/blob/main/src/lib/seam/components/index.ts

razor-x (commented):
Nice! We should probably add a story for the two components that render the lock button to check the layout renders ok when the button is not rendered.

xplato:
> Nice! We should probably add a story for the two components that render the lock button to check the layout renders ok when the button is not rendered.

@razor-x Sounds good to me! Is the preferred approach here just to add a new render function in the story like `WithoutLockUnlockButton` or something?

dawnho (approved):

razor-x:
Yea, we have been doing like this https://github.com/seamapi/react/blob/fe8d378a41f25601c8e68a22099e857f5a49946a/src/lib/seam/components/SupportedDeviceTable/SupportedDevice.stories.tsx#L22

razor-x (approved):

Issue #173, opened 2023-05-25
DeviceTable - Add account offline filter to health bar


mikewuu:
Spun off from #120 , the `<DeviceHealthBar/>` filter should also include a filter for offline accounts.

<img width="885" alt="Screen Shot 2023-06-28 at 2 19 09 PM" src="https://github.com/seamapi/react/assets/1843707/b0917d18-8398-4947-a5b1-ce0e3a6cc43f">

![CleanShot 2023-05-25 at 12 17 58](https://github.com/seamapi/react/assets/11449462/ded426ed-773e-4823-8968-0b99b58ff9d8)

[Figma](https://www.figma.com/file/Su3VO6yupz4yxe88fv0Uqa/Prebuilt-Components?type=design&node-id=521-60691&t=kwlzrmQ94iPjNbd7-0)

PR #255, opened 2023-06-28
feat: Add account issues filter to DeviceTable

dawnho:
Closes: https://github.com/seamapi/react/issues/173

Screen Shot 2023-06-28 at 3 36 22 PM

Added Account Offline status to device details:
Screen Shot 2023-06-28 at 4 38 02 PM


.storybook/seed-fake.js
```diff
@@ -204,13 +205,20 @@ export const seedFake = (db) => {
       supported_code_lengths: [6],
       name: 'GARAGE',
       battery: {
-        level: 0.9999532347993827,
+        level: 0.3,
         status: 'full',
```
razor-x, line 209:
```suggestion
        level: 0.29,
        status: 'low',
```

src/lib/seam/components/DeviceTable/DeviceRow.tsx
```diff
@@ -16,6 +17,48 @@ interface Props {
...
+    )
+  }
+  return (
+    <div className='seam-device-statuses'>
+      <DeviceOnlineStatus device={device} />
+      <BatteryStatus device={device} />
+      <LockStatus device={device} />
+    </div>
+  )
+}
```
razor-x, line 60: Few general style note for this project that we can't really lint against yet. (1) For non-component functions we prefer arrow functions and (2) bool consts and bool functions should be named with the `is`, etc., in front, e.g., `const isAccountOffline = ` and (3) these internal methods and components should be ordered after they are used, so they should all be below the main export, ref https://stackoverflow.com/questions/1760850/best-practice-ordering-of-public-protected-private-within-the-class-definition/1760877#1760877

src/lib/seam/components/DeviceTable/DeviceRow.tsx
```diff
@@ -25,22 +68,24 @@ export function DeviceRow({
   device,
   onClick,
 }: DeviceRowProps): JSX.Element | null {
+  const opacityClass =
```
razor-x, line 71: We should use `classNames` below and avoid the empty string. Then I think we can do like this

```
const opacityClassName = accountIsOffline(device) || deviceIsOffline(device) ? 'seam-offline-fade' ? undefined
```

and below

```
      <TableCell className={classNames('seam-table-cell', opacityClassName)}>

```

razor-x (approved):
Nice!! Just some code style nits.

PR #257, opened 2023-06-29
feat: Support React 16 and 17

razor-x:
Tested in a Vite app on React 16 and 17. Components seems to work fine with no errors. There are ways to run the Jest test suite on multiple versions of react, but we would need to expand the tests first for that to be useful.


dawnho (approved):
LGTM!

xplato (approved):

seveibar (approved):

PR #268, opened 2023-07-06
Update to prettier v3

razor-x:
- ci: Update to prettier v3
- build: Update to typescript v5.1

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -2,5 +2,6 @@
   "semi": false,
   "singleQuote": true,
   "jsxSingleQuote": true,
+  "trailingComma": "es5",
   "endOfLine": "lf"
 }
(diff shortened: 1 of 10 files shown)


.prettierrc.json
```diff
@@ -2,5 +2,6 @@
   "semi": false,
   "singleQuote": true,
   "jsxSingleQuote": true,
+  "trailingComma": "all",
```
razor-x, line 5: This is the new prettier v3 default, but it has to be set here because the icon generator uses the old default.

xplato (approved):

razor-x:
@mikewuu I don't want to merge this if it will mess up your draft PR. I know you are going ti break it up some though, so let me know what you prefer.

dawnho (approved):

Issue #259, opened 2023-06-29
Filter button position broken on Firefox


razor-x:
Screenshot is after clicking the filter button, the filter window far away to the bottom right.
![image](https://github.com/seamapi/react/assets/721372/94a1bd4d-aa03-47b5-8cad-fedf6d5aae9a)

razor-x:
This is only an issue in the Stroybook. Perhaps due to the story iframe and the `getBoundingClientRect` calculations. If that's the case, it might be broken for users that also render inside an iframe.

razor-x:
Closing in favor of https://github.com/seamapi/react/issues/424

PR #282, opened 2023-07-14
Add ability to set time-based codes

### LINKED ISSUE #236: Add support for Create / Edit / Delete Access Codes
# Tasks:
#### `<AccessCodeForm />`:
- [ ] Create the "Access Code Form" component

#### `<AccessCodeTable />`:
- [ ] Add `disableCreate`, `disableEdit`, `disableDelete` props to the `AccessCodesTable` component
- [ ] Show `+` button when `disableCreate` is `false`
- [ ] Show "menu" button on Code row when `disableEdit` and `disableDelete` is `false`

#### `<AccessCodeDetails />`:
- [ ] Add `disableEdit` and `disableDelete` props to the `AccessCodeDetails` component

# Designs

### `<AccessCodeForm />`
```
interface AccessCodeFormProps {
    accessCodeId?: string        // when set, this component is used for editing
    onCreate?: (ac: AccessCode) => void
    onEdit?: (ac: AccessCode) => void
}
```

Screen Shot 2023-06-30 at 5 08 40 PM

[Link to figma](https://www.figma.com/file/Su3VO6yupz4yxe88fv0Uqa/Seam-Components?type=design&node-id=895-82543&mode=design&t=QauFhLYXHZwRHBJv-0)

Timezone for the start and end time will be set to local time by default.

We can use a lightweight date and time pickers library for selecting time, will have to see if we can inject custom timezones with it. Open to suggestions on what to use here.

### `<AccessCodeDetails />`:
Image

[Link to figma](https://www.figma.com/file/Su3VO6yupz4yxe88fv0Uqa/Seam-Components?type=design&node-id=842-76549&mode=design&t=QauFhLYXHZwRHBJv-0)

### `<AccessCodeTable />`:

Image

[Link to figma](https://www.figma.com/file/Su3VO6yupz4yxe88fv0Uqa/Seam-Components?type=design&node-id=895-80211&mode=design&t=QauFhLYXHZwRHBJv-0)

Image

[Link to figma](https://www.figma.com/file/Su3VO6yupz4yxe88fv0Uqa/Seam-Components?type=design&node-id=844-79041&mode=design&t=QauFhLYXHZwRHBJv-0)

mikewuu:
Partially closes #236
Spin off from #269 

### Screenshots

https://github.com/seamapi/react/assets/11449462/0956f986-e20b-4322-b38f-ffb9ce2ee840


src/styles/_date-picker.scss
razor-x, line 1: Instead of checking this file in like this (which we would need to remember to update when we update flatpickr), can we somehow import directly? If not, maybe we need a code gen step to generate this file.

razor-x: Let's look at whats in the pkg https://www.npmjs.com/package/flatpickr?activeTab=code

mikewuu: I agree, copying-pasting the source like this isn't ideal. 

Unfortunately, they only export .css, and I couldn't get nested imports to work with a custom class, `.seam`. I wanted to nest it to avoid any of our style customizations from overriding the consumers (in case they're using flatpickr).

Internally they use [styl](https://github.com/tj/styl), which I don't see a way for us to hook into since we're using scss.

We could use https://github.com/Pikaday/Pikaday, which has scss, and so it should be easier for us to import / override.

razor-x: Where did the scss file come from? If we can generate it from styl or css we can add that to the build step.

src/styles/_time-picker.scss
```diff
@@ -0,0 +1,39 @@
...
+    z-index: 5000;
+    display: none;
+    position: fixed;
+    margin-top: 5px;
+    max-height: 230px;
+    overflow: auto;
+    padding: 8px 0;
+    width: 168px;
+    box-shadow:
+      0 2px 16px 2px rgb(15 22 28 / 15%),
```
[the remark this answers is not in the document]
mikewuu: Usually only used for shadows. Good to leave IMO.

src/styles/_date-picker.scss
mikewuu: Was copy-pasted manually. But here's my attempt at auto-generating it: https://github.com/seamapi/react/pull/282/commits/d34705f4514d1cf70929afaf8bbad9cbd22f76bb

Specifically in `npm run build:flatpickr-styles`:

```
echo \"// AUTO-GENERATED FILE DO NOT EDIT\n// See 'build:flatpickr-styles' for more info\n@mixin all {\n$(npx stylus --print src/styles/_date-picker-flatpickr.styl)\n}\" | sed -r 's/7ch.*;/7ch;/g' > src/styles/_date-picker-flatpickr.scss
```

- **_date-picker-flatpickr.styl** Add custom flatpickr theme that we're checking in.
- Output compiled stylus -> css
- Wrap in `@mixin all` to allow scoping in scss
- `sed -r 's/7ch.*;/7ch;/g'` - this is to remove a weird character that stylus outputted. It might exist in the original source code. It breaks sass build though, so it had to go.
- output to **_date-picker-flatpickr.scss**
- Finally, import the generated scss in **_date-picker.scss**

CleanShot 2023-07-17 at 22 35 23

razor-x: Ok, this is better, but still awkward as it relies on sed. I might take a stab at this.

Looking at the lib options, it's unclear if they are still even maintained. Would it be crazy to use native date pickers? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

mikewuu: Definitely wouldn't mind going with the native picker as long as the lack of styles is ok.

@dawnho ?

package.json
```diff
@@ -77,7 +80,7 @@
     "test:update": "jest --updateSnapshot",
     "test:watch": "jest --watch",
     "test:inspect": "node --inspect-brk ./node_modules/jest/bin/jest.js --runInBand",
-    "lint": "eslint --ignore-path .gitignore --ext .js,.cjs,.mjs,.ts,.tsx .",
+    "lint": "eslint --ignore-path .gitignore --ext .js,.cjs,t.mjs,.ts,.tsx .",
```
razor-x, line 83: Looks like a mistake. We should revert to avoid merge conflicts.
```suggestion
    "lint": "eslint --ignore-path .gitignore --ext .js,.cjs,.mjs,.ts,.tsx .",
```

razor-x (changes requested):
First round of comments limited to some minor things. This is great work so far, really the time and date stuff is non-trivial.

This PR will introduce two new areas of complexity I would like to manage before we merge. I think it's worth getting this right now since we will keep building on it.

1. flatpickr. I think it still may be the best choice at this point, but I want to see if I can find a solution that does not require checking in the scss file.
2. I need more time to understand the date manipulation. So far the top level component APIs looks good, but I think we should leverage luxon more. I want to avoid string manipulation wherever possible. The lib files are a great target for unit testing since it's hard to represent all cases in the storybook.

mikewuu:
@razor-x updated to use native pickers in https://github.com/seamapi/react/pull/282/commits/ecf1aeb738136b67b0334a3144dadc4ef0c68546, and https://github.com/seamapi/react/pull/282/commits/5f1f861b1d41381cccd3f928f1438899911a54d4.

Other than the obvious lack of styling, here are a couple other missing features from the design:

- Can't force time picker to show / hide seconds, it seems to be tied to the browser's preference.
- Can't modify time picker intervals. There's a step option but that's the increment/decrement size when clicking on arrows, not the option intervals.

https://github.com/seamapi/react/assets/11449462/cf675719-e135-401b-8812-67c6b29dc95b

package.json
```diff
@@ -177,6 +177,7 @@
     "storybook-addon-designs": "^7.0.0-beta.2",
     "stylelint": "^15.6.0",
     "stylelint-config-standard-scss": "^10.0.0",
+    "stylus": "^0.59.0",
```
razor-x, line 180:
```suggestion
```

src/lib/ui/DateTimePicker/DateTimePicker.tsx
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    const updatedValue = `${rawDate}T${rawTime}`
+    onChange(updatedValue)
+  }, [rawDate, rawTime, onChange, date, time])
+
+  return (
+    <div className={classNames('seam-date-time-picker', className)}>
+      <DatePicker value={rawDate} onChange={setRawDate} {...props} ref={ref} />
+      <TimePicker value={rawTime} onChange={setRawTime} {...props} />
```
razor-x, line 51: @mikewuu Should we use https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local instead of having them separated?

src/lib/seam/components/AccessCodeForm/TimezonePicker/TimezonePicker.tsx
```diff
@@ -0,0 +1,80 @@
...
+          className='seam-manual-timezone-checkbox'
+        />
+        <div
+          className={classNames('seam-timezones', {
+            'seam-disabled': !isManualTimezone,
+          })}
+        >
+          <ul>
+            {getTimezones()
+              .sort(
```
razor-x, line 55: Let's move this to [dates.ts](https://github.com/seamapi/react/blob/main/src/lib/dates.ts) as `compareByTimezoneOffsetAsc`

src/lib/seam/components/AccessCodeForm/TimezonePicker/TimezonePickerListItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+      return
+    }
+
+    const { current: item } = elRef
+
+    if (item == null) {
+      return
+    }
+
+    const container = item.closest('.seam-timezones')
```
razor-x, line 30: Interesting, can we pass the container as a prop instead of having to search by class name?

src/lib/seam/components/AccessCodeForm/TimezonePicker/TimezonePickerListItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    if (container == null) {
+      return
+    }
+
+    const { height: containerHeight } = container.getBoundingClientRect()
```
razor-x, line 36: This looks similar to the logic we use to position the filter box, but we have this bug https://github.com/seamapi/react/issues/259

Can we use native `<select>` instead?

razor-x (approved):
We will go with the native pickers for now. The style is fine and have higher compatibility with less complexity. If we find gaps with the native, we will treat a custom date picker as a separate feature and do a comprehensive evaluation. We may tweak the native picker options more, but again, that can be something we do after this is merged.

Going to unblock merging this as the remaining suggestions are mostly internal details that can be addressed in future PRs. But we should resolve conversions or move to issues before merge.

src/lib/ui/RadioField/RadioField.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+interface RadioFieldContextProps<Value extends string> {
+  value: Value
+  onChange: (value: Value) => void
+  name: string
+}
+
+const RadioFieldContext = createContext<
+  RadioFieldContextProps<any> | undefined
```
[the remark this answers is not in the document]
mikewuu: Not that I'm aware of, but this `any` might be ok, because:

- The generic is be specified via `useRadioField`.
- The `RadioFieldContext` isn't actually exposed (exported). The `any` is self-contained.
- The single call-site, inside `<Radio/>` forwards the type along:

```ts

export function Radio<Value extends string>({
  value,
  label,
}: RadioProps<Value>): JSX.Element {
  const { value: fieldValue, onChange, name } = useRadioField<Value>()
...
}
```

Ultimately, we still get strong typing where it matters, in the primary component: `RadioField`


CleanShot 2023-07-20 at 15 08 37

We've passed in a union for the `value`, and...

CleanShot 2023-07-20 at 15 09 33

The options are correctly inferred.

src/lib/ui/FormField.tsx
```diff
@@ -38,7 +40,11 @@ export function FormField({
       })
     }
 
-    if (component.type === TextField) {
+    if (
```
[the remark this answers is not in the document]
mikewuu: Not required IMO. Any unknown components are just forwarded along, unchanged.

> Does this component work if more than two children are passed in?

Yep, it's just a `map` so it can accept any number of elements. 

https://github.com/seamapi/react/assets/11449462/a8a554cb-e248-4122-bb91-c6732b5d2bb7

All the labels get the same click binding!

src/lib/ui/RadioField/RadioField.tsx
```diff
@@ -0,0 +1,58 @@
...
+export interface RadioOption<Value> {
+  label: string
+  value: Value
+}
+
+interface RadioFieldContextProps<Value extends string> {
+  value: Value
+  onChange: (value: Value) => void
+  name: string
+}
```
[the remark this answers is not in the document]
mikewuu: ahh, yep. Updated in https://github.com/seamapi/react/pull/282/commits/8b0c10a55327ac9d2d32ebbee11df53272172614

src/lib/seam/components/AccessCodeForm/AccessCodeForm.tsx
```diff
@@ -54,6 +74,24 @@ function Content({
       return
     }
 
+    if (type === 'time_bound') {
```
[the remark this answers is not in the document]
mikewuu: Yep, good idea. https://github.com/seamapi/react/pull/282/commits/9ffd1d80f99f390cd43418593283e4680df6f397

src/lib/seam/components/AccessCodeForm/AccessCodeForm.tsx
```diff
@@ -93,6 +157,48 @@ function Content({
...
+              },
+              {
+                label: t.typeTimeBoundLabel,
+                value: 'time_bound',
+              },
+            ]}
+          />
+          <>
+            {type === 'time_bound' && (
+              <div className='seam-selected-times'>
```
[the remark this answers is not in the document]
mikewuu: Sure, updated in https://github.com/seamapi/react/pull/282/commits/c3342e12e8da035cb5473527117a9e0ec87e14a7

src/lib/seam/components/AccessCodeForm/TimezonePicker/TimezonePicker.tsx
mikewuu: Added in https://github.com/seamapi/react/pull/282/commits/3f987acb610ff1e28b52c96caacc4b18140d0935

src/lib/ui/DateTimePicker/DateTimePicker.tsx
mikewuu: Updated inhttps://github.com/seamapi/react/pull/282/commits/59360b464ae3e77cf486b0084455e975c7653d00

src/lib/seam/components/AccessCodeForm/TimezonePicker/TimezonePickerListItem.tsx
mikewuu: Sure, added `TimezonPicker` https://github.com/seamapi/react/pull/282/files#diff-66e36b728600047d9d211deedbdd8186d978821adc5bc075e46c35e55baa7068

src/lib/seam/components/AccessCodeForm/TimezonePicker/TimezonePickerListItem.tsx
mikewuu: No longer relevant as using native `select`.

src/lib/dates.ts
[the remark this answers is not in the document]
mikewuu: Yep. When we update them, the tests in https://github.com/seamapi/react/pull/282/commits/ed544f01b444071ee3dfbaa66991272a63da7a58 should confirm no regressions.

mikewuu:
@razor-x I believe all threads have been addressed. Let me know if I've missed any!

dawnho (approved):
LGTM!

the timezone field looks like it can use a search bar, but that can be a later later improvement

razor-x:
Nice! I need to test a few things first, but I'll plan to merge this today!

src/lib/ui/DateTimePicker/DateTimePicker.tsx
razor-x: We may need to go back to keeping them separate. Firefox support is no good for this pne. It doesn't work on my desktop. It only shows the date picker.

PR #295, opened 2023-07-21
feat(AccessCodeForm): invert set time zone manually checkbox

mikewuu:
Requested via [Slack]( https://hello-seam.slack.com/archives/C051ETBPRTQ/p1689874525095959?thread_ts=1689808685.428099&cid=C051ETBPRTQ)

https://github.com/seamapi/react/assets/11449462/0c3c7a1a-a515-4f83-8610-11f2cfe6ebf1

### DIFF
--- a/src/lib/seam/components/AccessCodeForm/AccessCodeFormTimezonePicker.tsx
+++ b/src/lib/seam/components/AccessCodeForm/AccessCodeFormTimezonePicker.tsx
@@ -41,8 +41,10 @@ export function AccessCodeFormTimezonePicker({
       <div className='seam-content'>
         <Checkbox
           label={t.setTimezoneManuallyLabel}
-          checked={isManualTimezone}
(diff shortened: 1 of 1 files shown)


razor-x:
Can you make it so when we click on the select box, if the box is not checked, it checks the box and opens the selection up?

mikewuu:
I think so, but it might be better to have user take explicit action (toggling) box instead so that they're aware of the switch from manual/auto. wdyt?

razor-x:
IMO this is already better. Personally I want to click the box right away and not be slowed down by an extra click, but I can be convinced the extra click is needed.

CC @miken-seam

razor-x (approved):
I like it. Let's hold off merging until we get some more feedback in the comments.

dawnho (approved):
I agree with @razor-x 's suggestion about opening up the list when the checkbox is unchecked

PR #296, opened 2023-07-21
Update fake-seam-connect

razor-x:
- ci: Do not test on macos or windows
- test: Update fake-seam-connect


razor-x:
I think we are good to bump fake-seam-connect to v1 before merge.

seveibar (approved):

PR #299, opened 2023-07-21
Update to Storybook v7.1

razor-x:
- build: Update to storybook v7.1
- Fix type error with defineConfig


package.json
```diff
@@ -175,7 +176,6 @@
     "react-dom": "^18.2.0",
     "sass": "^1.62.1",
     "storybook": "^7.0.2",
```
xplato, line 178: Is this supposed to be `^7.1.x` or is it just the `@storybook/addon-designs` package that needed to be updated?

razor-x: Yea just the design package. The other updates are in the lockfile. In general, updating versions in the package.json is not needed if an `npm update` is ok.

xplato (approved):

PR #301, opened 2023-07-23
add endpoint to provider props to enable endpoint override for web components

seveibar:


dawnho:
What is this for?

seveibar:
@dawnho This allows you to specify an endpoint in webcomponents which is required to have it work on the documentation site, which uses fake seam connect at https://react.seam.co/api

xplato (approved):

PR #304, opened 2023-07-24
Process empty string endpoint as "endpoint not provided", refactor SeamProvider for readability

seveibar:


src/lib/seam/SeamProvider.tsx
```diff
@@ -70,42 +70,40 @@ export function SeamProvider({
...
-  const value = useMemo(() => {
+  const seamProviderContextValue = useMemo(() => {
     const context = createSeamContextValue(props)
     if (
       context.client == null &&
       context.publishableKey == null &&
       context.clientSessionToken == null
     ) {
-      return defaultSeamContextValue
+      return getDefaultSeamContextValue()
```
seveibar, line 80: This behavior is slightly different but more correct. If the `globalThis.seam` client is defined after this file is imported it will now correctly grab it

src/lib/seam/use-seam-client.ts
```diff
@@ -18,6 +18,10 @@ export function useSeamClient(): {
     clientSessionToken,
     ...context
   } = useSeamContext()
+
+  // eslint-disable-next-line
+  const endpoint = clientOptions?.endpoint || "https://connect.getseam.com"
```
seveibar: this is the actual fix, web components are now defining `endpoint` so adds the default. You could alternatively do this inside of seamapi/javascript

razor-x: The source of truth for the default endpoint should be in `seamapi`. I think this may be an upstream issue in seamapi, why does `endpoint: undefined` not result in the default endpoint used by the `Seam` client?

seveibar: i think this is most likely the "real issue" upstream

image

src/lib/seam/SeamProvider.tsx
```diff
@@ -70,42 +70,42 @@ export function SeamProvider({
...
-
   return (
     <div className='seam-components'>
       <QueryClientProvider
         client={queryClient ?? globalThis.seamQueryClient ?? defaultQueryClient}
       >
-        <Provider value={value}>{children}</Provider>
+        <seamContext.Provider value={seamProviderContextValue}>
+          {children}
+        </seamContext.Provider>
```
razor-x, line 102: This matches the existing naming convention in this project when working with providers.

```suggestion
        <Provider value={seamProviderContextValue}>
          {children}
        </Provider>
```

src/lib/seam/use-seam-client.ts
seveibar: image

src/lib/seam/SeamProvider.tsx
razor-x: Cool. Can you undo the variable rename though, or simply name it `contextValue`? I do not think the longer name is needed in this context.

src/lib/seam/SeamProvider.tsx
```diff
@@ -145,9 +145,9 @@ const createSeamContextValue = (options: SeamProviderProps): SeamContext => {
   return { client: null }
 }
 
-const defaultSeamContextValue = createDefaultSeamContextValue()
-
-export const seamContext = createContext<SeamContext>(defaultSeamContextValue)
+export const seamContext = createContext<SeamContext>(
+  getDefaultSeamContextValue()
+)
```
razor-x, line 150:
```suggestion
const defaultSeamContextValue = createDefaultSeamContextValue()

export const seamContext = createContext<SeamContext>(defaultSeamContextValue)
```

src/lib/seam/use-seam-client.ts
razor-x: Yea, I would consider this an issue with seamapi. When undefined is treated differently then "missing" it can cause surprising behavior. IMO seamapi should be updated (in a major version) to unify that behavior.

To fix the issue in this project, we should omit the option instead of redefine the default.

src/lib/seam/SeamProvider.tsx
seveibar: I'm probably going to close this PR and go upstream to seamapi (fair), but this convention of having ambiguous "value" and "Provider" thrown around isn't a good naming pattern I think- `value`, like `data` shouldn't be a var. If the file was less complex it might be passable but this file is really complex

src/lib/seam/use-seam-client.ts
razor-x: See also here, this should really be `apiKey: null` (I think the types do not allow that though, hence the `undefined`).

PR #309, opened 2023-07-24
feat: Add `title` prop for table components

### LINKED ISSUE #308: Add `title?: string | null` prop to table components with top-level title

xplato:
Closes #308 • Adds a `title` prop to table components with a top-level title.

### DIFF
--- a/src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
+++ b/src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
@@ -51,6 +51,7 @@ export interface AccessCodeTableProps {
   onAccessCodeClick?: (accessCodeId: string) => void
   preventDefaultOnAccessCodeClick?: boolean
   onBack?: () => void
+  title?: string | null
   className?: string
 }
 
@@ -78,6 +79,7 @@ export function AccessCodeTable({
   onBack,
   accessCodeFilter = defaultAccessCodeFilter,
   accessCodeComparator = compareByCreatedAtDesc,
+  title = t.accessCodes,
   className,
 }: AccessCodeTableProps): JSX.Element {
   const { accessCodes } = useAccessCodes({
@@ -140,9 +142,13 @@ export function AccessCodeTable({
       <ContentHeader onBack={onBack} />
       <TableHeader>
         <div className='seam-left'>
-          <TableTitle>
-            {t.accessCodes} <Caption>({filteredAccessCodes.length})</Caption>{' '}
-          </TableTitle>
+          {title != null ? (
+            <TableTitle>
+              {t.accessCodes} <Caption>({filteredAccessCodes.length})</Caption>{' '}
+            </TableTitle>
+          ) : (
+            <div className='seam-fragment' />
+          )}
           {!disableCreateAccessCode && (
             <IconButton
               onClick={toggleAddAccessCodeForm}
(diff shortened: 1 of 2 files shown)


xplato:
@razor-x is this approach what you had in mind?

src/lib/seam/components/DeviceTable/DeviceTable.tsx
```diff
@@ -109,9 +111,13 @@ export function DeviceTable({
...
       <TableHeader>
-        <TableTitle>
-          {t.devices} <Caption>({filteredDevices.length})</Caption>
-        </TableTitle>
+        {title != null ? (
+          <TableTitle>
+            {title ?? t.devices} <Caption>({filteredDevices.length})</Caption>
+          </TableTitle>
+        ) : (
+          <div />
```
xplato, line 119: Not sure if we want this. Preserves the location of the search field.

razor-x: Ah, yea good catch on the layout quirk. I don't love an empty div as a solution though, should be via class name if possible 🤔

xplato: Using `<div className='seam-fragment' />` for now. If you want a different className and/or extracted to a atomic component, let me know!

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
```diff
@@ -140,9 +142,13 @@ export function AccessCodeTable({
       <ContentHeader onBack={onBack} />
       <TableHeader>
         <div className='seam-left'>
-          <TableTitle>
-            {t.accessCodes} <Caption>({filteredAccessCodes.length})</Caption>{' '}
-          </TableTitle>
+          {title != null ? (
+            <TableTitle>
+              {t.accessCodes} <Caption>({filteredAccessCodes.length})</Caption>{' '}
```
xplato, line 147: Do we want to add another prop (in a separate PR) for hiding the device count as well?

razor-x: Interesting so if they override the title then there is no way to see device count. I think for this first version it's ok.

We could either allow the title prop to be a template literal that allows them to specify a placeholder for the count, or we could add a new prop onDevices

xplato: > Interesting so if they override the title then there is no way to see device count. I think for this first version it's ok.

The opposite, actually. They can always see the device count unless `title === null`. I was just thinking ahead that some people _might_ wanna remove the count. We can wait until a customer asks for this though

razor-x: Exposing data related hooks as props probably needs more careful thought. I don't think we should add it adhoc here.

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
```diff
@@ -140,9 +142,14 @@ export function AccessCodeTable({
       <ContentHeader onBack={onBack} />
       <TableHeader>
         <div className='seam-left'>
-          <TableTitle>
-            {t.accessCodes} <Caption>({filteredAccessCodes.length})</Caption>{' '}
-          </TableTitle>
+          {title != null ? (
+            <TableTitle>
+              {title ?? t.accessCodes}{' '}
+              <Caption>({filteredAccessCodes.length})</Caption>{' '}
```
razor-x, line 148: This nbsp is new not in the other one. Is it intentional?
```suggestion
              <Caption>({filteredAccessCodes.length})</Caption>
```

razor-x (approved):
Let's update the docs after merge

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
xplato: Sounds good to me! We can push this to an issue 👍

PR #313, opened 2023-07-24
feat: Add battery percentage to `BatteryStatus`

Andrewjeska:
Customer request:

https://hello-seam.slack.com/archives/C04SGND5QL9/p1690220694132189?thread_ts=1690220662.657769&cid=C04SGND5QL9

Screenshot 2023-07-24 at 3 45 03 PM

### DIFF
--- a/src/lib/ui/device/BatteryStatus.tsx
+++ b/src/lib/ui/device/BatteryStatus.tsx
@@ -26,7 +26,7 @@ function Content(props: {
 }): JSX.Element | null {
   const { status, level } = props
 
-  const percentage = level ? ` (${Math.floor(level * 100)}%)` : null
+  const percentage = level != null ? ` (${Math.floor(level * 100)}%)` : null
 
   if (status === 'full') {
     return (
--- a/src/lib/ui/device/BatteryStatus.tsx
+++ b/src/lib/ui/device/BatteryStatus.tsx
@@ -1,4 +1,4 @@
-import type { BatteryStatus as SeamBatteryStatus, CommonDevice } from 'seamapi'
+import type { CommonDevice, BatteryStatus as SeamBatteryStatus } from 'seamapi'
 
 import { BatteryLevelCriticalIcon } from 'lib/icons/BatteryLevelCritical.js'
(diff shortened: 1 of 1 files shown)


src/lib/ui/device/BatteryStatus.tsx
```diff
@@ -12,21 +12,27 @@ interface BatteryStatusProps {
...
+        status={props.device.properties.battery?.status}
+        level={props.device.properties.battery?.level}
+      />
     </div>
   )
 }
 
 function Content(props: {
   status: SeamBatteryStatus | null | undefined
+  level: number | undefined
```
razor-x, line 25: If this is allowed undefined then we need to update the string logic so it doesn't render the word undefined.

razor-x, line 25:
```suggestion
  level: number | null | undefined
```

Andrewjeska: [9b2cbd9](https://github.com/seamapi/react/pull/313/commits/9b2cbd9950f4b3248602fe422921c07d7f537cde)

Andrewjeska: [72eafc6](https://github.com/seamapi/react/pull/313/commits/72eafc69938ed23f018c3367dcbbb7df40b2af74)

src/lib/ui/device/BatteryStatus.tsx
```diff
@@ -12,21 +12,29 @@ interface BatteryStatusProps {
...
 }
 
 function Content(props: {
   status: SeamBatteryStatus | null | undefined
+  level: number | null | undefined
 }): JSX.Element | null {
-  const { status } = props
+  const { status, level } = props
+
+  const percentage = level ? ` (${Math.floor(level * 100)}%)` : null
```
razor-x, line 29: The linter error is protecting us from rendering null instead of 0 in the 0 level case.

```suggestion
  const percentage = level == null ? ` (${Math.floor(level * 100)}%)` : null
```

razor-x (approved):

PR #314, opened 2023-07-24
`useClimateSettingSchedules` hook

Andrewjeska:


src/lib/seam/thermostats/climate-setting-schedules/use-climate-setting-schedules.ts
```diff
@@ -0,0 +1,53 @@
...
+              climate_setting_schedule_id: schedule.climate_setting_schedule_id,
+            },
+          ],
+          schedule
+        )
+      }
+    },
+  })
+
+  return { ...rest, climate_setting_schedules: data }
```
razor-x, line 52: Should be `climateSettingSchedules`. See the other multi word hook.

razor-x (changes requested):
Looks good, but one blocking fix

src/lib/seam/thermostats/climate-setting-schedules/use-climate-setting-schedules.ts
Andrewjeska: [69d3fd0](https://github.com/seamapi/react/pull/314/commits/69d3fd0bf6b944c009051ccbefebd2c0f7e85c6d)

razor-x (approved):

PR #323, opened 2023-07-28
fix: Update to seamapi v8.2.2

razor-x:


seveibar (approved):

razor-x:
Something is mis-confiruged, GET requests are returning "CORS Missing Allow Credentials". API may need to return `access-control-allow-credentials` on other methods (it only returns it on OPTIONS now?)

This option should mean the client side is allowed to make the request https://github.com/seamapi/javascript/blob/62dbfb9c87fdc9314045328ac51fa8d0b8ca1043/src/seam-connect/client.ts#L96

PR #331, opened 2023-08-01
feat: Accept csv or json for web component array attributes

razor-x:
- feat: Accept csv or json for web component array attributes
- Run npm update

See https://github.com/rxfork/react-to-web-component/pull/3

### DIFF
--- a/src/lib/seam/components/SupportedDeviceTable/SupportedDeviceTable.element.ts
+++ b/src/lib/seam/components/SupportedDeviceTable/SupportedDeviceTable.element.ts
@@ -7,7 +7,7 @@ export const name = 'seam-supported-device-table'
 export const props: ElementProps<SupportedDeviceTableProps> = {
   cannotFilter: 'boolean',
   disableFilter: 'boolean',
-  brands: 'json',
+  brands: 'array',
   className: 'string',
 }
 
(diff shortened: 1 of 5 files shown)


dawnho (approved):

src/lib/seam/components/SupportedDeviceTable/SupportedDeviceTable.element.ts
```diff
@@ -7,7 +7,7 @@ export const name = 'seam-supported-device-table'
 export const props: ElementProps<SupportedDeviceTableProps> = {
   cannotFilter: 'boolean',
   disableFilter: 'boolean',
-  brands: 'json',
+  brands: 'array',
```
ethanplee14, line 10: I'm not familiar with the rest of the code base but is this all we need for it to be parsed properly? I believe we parsed brand by using `JSON.parse(brands)`. I assume this changes the flow to a new means of parsing which allows for CSV? 

If that's the case, does this break the old behavior?

ethanplee14: Sorry I didn't see the link to `react-to-web-components`

ethanplee14 (approved):
Thank you!

src/lib/seam/components/SupportedDeviceTable/SupportedDeviceTable.element.ts
razor-x: Yea, I forked the lib we are using to support our requirements. I may try to PR back upstream, but for now this will keep things moving for us.

PR #333, opened 2023-08-03
feat(SupportedDeviceTable): Add excludedBrands prop

razor-x:


src/lib/seam/components/SupportedDeviceTable/SupportedDeviceContent.tsx
```diff
@@ -14,19 +14,22 @@ interface SupportedDeviceContentProps {
   resetFilterValue: () => void
   filters: DeviceModelFilters
   brands: string[] | null
+  omitBrands: string[]
```
mikewuu, line 17: What about `disabledBrands`, or `hiddenBrands`?

mikewuu (approved):

dawnho (approved):
Good idea!

src/lib/seam/components/SupportedDeviceTable/SupportedDeviceContent.tsx
dawnho: I like `hideBrands`

mikewuu: that still sounds verb-y though. 🤷‍♂️

razor-x: I considered both. It can't be disable since that would be a boolean props. It could be `disabledBrands`. Actually, omit is a verb, it should be omitted brands, but that's awkward.

Maybe `hiddenBrands`?

dawnho: `disabledBrands` is nice. We use the word 'disable' elsewhere, so it's nice to use familiar vocabulary

razor-x: Yes, but in this case `disable` and `disabled` is easy to mess up / misread (one letter), and disable is always used for bool props.

razor-x: @dawnho What about `excludedBrands`. I think include / exclude is a standard term.

xplato (approved):

PR #339, opened 2023-08-07
Add thermostat devices to `DeviceTable`

xplato:
Adds thermostat devices to the `DeviceTable`

image


razor-x:
Why does it show a lock icon / unknown?

xplato:
> Why does it show a lock icon / unknown?

@razor-x the `DeviceImage` component uses `device.properties.image_url`, but this is unspecified for thermostats in `seed-fake`. Looks like it is missing in `seam-connect`, too. I thought this could be merged in without release, but I can add those assets to `seam-connect` first.

PR #347, opened 2023-08-09
fix: Set CORS headers for assets

razor-x:
Needed to allow js files to be loaded via script tag. Verified working via 
image

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -72,6 +72,10 @@
         {
           "key": "Access-Control-Allow-Headers",
           "value": "*"
+        },
+        {
+          "key": "Cache-Control",
+          "value": "public, max-age=31536000, immutable"
         }
       ]
     },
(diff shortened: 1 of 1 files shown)


vercel.json
```diff
@@ -54,6 +54,31 @@
         }
       ]
     },
+    {
+      "source": "/v/(.*)",
+      "headers": [
+        {
+          "key": "Access-Control-Allow-Origin",
+          "value": "*"
```
seveibar, line 62: I'm skeptical this will work too well, iirc "*" doesn't work as well as returning the request origin

seveibar (approved):

vercel.json
razor-x: I think that is correct, although I cannot remember the cases where it might create an issue. Possibly with `localhost`. To further complicate things, the "rules" when disallowing credentials / for loading via `script` tag may be different than for fetch / XHR.

PR #353, opened 2023-08-16
Add ThermostatCard

xplato:
Adds the `ThermostatCard` component as part of the thermostat detail view

Screenshot 2023-08-21 at 10 23 01 AM


src/lib/ui/thermostat/ThermostatCard.tsx
```diff
@@ -0,0 +1,167 @@
...
+                    </>
+                  )}
+                </div>
+              </>
+            )}
+
+          {'current_climate_setting' in device.properties &&
+            'hvac_mode_setting' in
+              (device.properties
+                ?.current_climate_setting as ClimateSetting) && (
```
xplato, line 103: I don't love the conditions here. It would be nice to have some additional generic typing on `useDevice` but idk if we wanted to do that atm

razor-x: Noted. This is part of a larger pattern I am hoping we can solve by eventually moving to `@seamapi/types`. For now, you can add some type guards to `src/lib/seam/devices/types.ts` with there conditions so we can organize the type situation and make sure to capture the different type variants upstream.

(I am considering any files names `types.ts` as an anti-pattern indicating some gap we are shimming for our upstream types).

razor-x: Extending on that, instead of or in addition to the type guards, you can improve the nesting situation by just making components that return null when passed devices it can't render. 

See for example https://github.com/seamapi/react/blob/main/src/lib/ui/device/LockStatus.tsx

src/lib/ui/thermostat/ThermostatCard.tsx
```diff
@@ -0,0 +1,171 @@
...
+    <div className='seam-thermostat-card'>
+      <Content deviceId={deviceId} />
+    </div>
+  )
+}
+
+function Content(props: { deviceId: string }): JSX.Element | null {
+  const { deviceId } = props
+
+  const { device } = useDevice({
```
razor-x, line 27: I'm not sure ThermostatCard will own the loading state vs the parent component. I believe the things inside `ui` are presentational components that assume data loading is handled by the parent.  Maybe this should take a `device` prop instead of a device_id?

xplato: Sounds good to me. The reason I added `deviceId` initially was so I could easily use it in a story. How can I access specific devices from the seed-fake in a story?

razor-x: What if we just move the hook into the story?

xplato: Oh duh lol—I keep thinking stories are on a different plane or something. Fixed, thanks!

src/lib/ui/thermostat/ThermostatCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  const {
+    temperature_fahrenheit: temperatureFahrenheit,
+    temperature_celsius: temperatureCelsius,
+    current_climate_setting: currentClimateSetting,
+    is_fan_running: isFanRunning,
+    relative_humidity: relativeHumidity,
+
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
razor-x, line 50: Do we have an upstream issue with the SDK types?

src/lib/ui/thermostat/ThermostatCard.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+    const { device } = useDevice({
+      device_id: 'device5',
+    })
+
+    return (
+      <Box display='grid' gap={4} gridTemplateColumns='repeat(4, min-content)'>
+        <ThermostatCard device={device as ThermostatDevice} />
+      </Box>
+    )
+  },
```
razor-x, line 27:
```suggestion
  render: () => {
    const { device } = useDevice({
      device_id: 'device5',
    })
    
    if (!isThermostatDevice(device)) return null

    return (
      <Box display='grid' gap={4} gridTemplateColumns='repeat(4, min-content)'>
        <ThermostatCard device={device} />
      </Box>
    )
  },
```

src/lib/ui/thermostat/ThermostatCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+          </div>
+
+          <div className='seam-thermostat-property-block'>
+            <p className='seam-thermostat-property-label'>{t.systemStatus}:</p>
+          </div>
+          <div className='seam-thermostat-property-block'>
+            <div
+              className={classNames(
+                'seam-thermostat-property-tag',
+                `seam-thermostat-property-tag-${'cooling'}`
```
razor-x, line 125:
```suggestion
                `seam-thermostat-property-tag-cooling`
```

src/lib/ui/thermostat/ThermostatCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+}
+
+function Content(props: { device: ThermostatDevice }): JSX.Element | null {
+  const { device } = props
+
+  const [temperatureUnit, setTemperatureUnit] = useState<
+    'fahrenheit' | 'celsius'
+  >('fahrenheit')
+
+  const toggleTemperatureScale = (): void => {
```
razor-x, line 33: I wonder if `scale` would have been a better API term than unit 🤔 

```suggestion
  const toggleTemperatureUnit = (): void => {
```

src/lib/ui/thermostat/ThermostatCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+    setTemperatureUnit(
+      temperatureUnit === 'fahrenheit' ? 'celsius' : 'fahrenheit'
+    )
+  }
+
+  if (device == null || (device != null && !isThermostatDevice(device))) {
+    return null
+  }
+
```
razor-x, line 42: Do we need this check? It should be handled already by the types / parent.
```suggestion
```

src/lib/ui/thermostat/ThermostatCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+          <div className='seam-thermostat-property-block'>
+            <p className='seam-thermostat-property-value'>
+              {Math.trunc(
+                Number(
+                  temperatureUnit === 'fahrenheit'
+                    ? temperatureFahrenheit
+                    : temperatureCelsius
+                )
+              )}
+              º
```
razor-x, line 86:
```suggestion
```

src/lib/ui/thermostat/ThermostatCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+          </button>
+        </div>
+
+        <div className='seam-thermostat-properties'>
+          <div className='seam-thermostat-property-block'>
+            <p className='seam-thermostat-property-label'>{t.temperature}:</p>
+          </div>
+          <div className='seam-thermostat-property-block'>
+            <p className='seam-thermostat-property-value'>
+              {Math.trunc(
```
razor-x, line 79: Can we re-use this component? Maybe it needs a more general name? https://react.seam.co/?path=/docs/library-climatesettingstatus--docs

src/lib/ui/thermostat/ThermostatCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+                    : temperatureCelsius
+                )
+              )}
+              º
+            </p>
+
+            <p className='seam-thermostat-property-value'>|</p>
+            <p className='seam-thermostat-property-label'>{t.humidity}:</p>
+            <p className='seam-thermostat-property-value'>
+              {Number(relativeHumidity) * 100}%
```
razor-x, line 92: Is relativeHumidity a string on the API?

src/lib/ui/thermostat/ThermostatCard.tsx
xplato: Yeah, I think so. I'm not too sure what the underlying issue is. Basically, we need `ThermostatDevice` to have `properties` typed as `ThermostatDeviceProperties` (which inherits from `CommonDeviceProperties`), instead of `CommonDeviceProperties` alone. Maybe it's an issue in this repo, not sure.

src/lib/ui/thermostat/ThermostatCard.stories.tsx
xplato: `isThermostatDevice` won't accept `null` or `undefined`, so I have to add some additional conditions here. (Should we update these functions to accept empty vals though?)

src/lib/ui/thermostat/ThermostatCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+          </div>
+        </div>
+      </div>
+    </div>
+  )
+}
+
+const t = {
+  fahrenheit: 'F˚',
+  celsius: 'C˚',
```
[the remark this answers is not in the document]
xplato: Oh thanks 🤦‍♂️

src/lib/ui/thermostat/ThermostatCard.tsx
xplato: Oops, that's old—thanks!

src/lib/ui/thermostat/ThermostatCard.tsx
xplato: This is used since the temperatures are raw numbers. I can add this to `t` though

src/lib/ui/thermostat/ThermostatCard.stories.tsx
razor-x: Ah I see. No let's not change them to allow null. In this case, we can just update the story with the missing loading / error states. Really for the story you could put

```
if (device == null) return null
```

If you want to get fancy you could add a MUI loading spinner since the stories are allowed to use MUI.

src/lib/ui/thermostat/ThermostatCard.tsx
xplato: Ah thanks—old from before type guard

src/lib/ui/thermostat/ThermostatCard.tsx
xplato: Are you just thinking of something like `Temperature` with F, C, and unit props?

razor-x: Sure, Temperature sounds nice!

src/lib/ui/thermostat/ThermostatCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+  off: 'Off',
+  temperature: 'Temperature',
+  humidity: 'Humidity',
+  setting: 'Setting',
+  fanMode: 'Fan mode',
+  systemStatus: 'System status',
+  cooling: 'Cooling',
+  heating: 'Heating',
+  fan: 'Fan only',
+  degree: 'º',
```
razor-x, line 154:
```suggestion
```

src/lib/ui/thermostat/ThermostatCard.tsx
razor-x: Ok. I can look into this one. Ok to have the work around in for now to unblock.

src/lib/ui/thermostat/ThermostatCard.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+type Story = StoryObj<typeof ThermostatCard>
+
+export const Content: Story = {
+  render: () => {
+    const { device } = useDevice({
+      device_id: 'device5',
+    })
+
+    if (device == null) return <></>
+
```
razor-x, line 22:
```suggestion
    if (device == null) return null
    if (!isThermostatDevice(device)) return null
```

razor-x (approved):
Approving to unblock, but I noticed two minor things we should fix:

1. There is [CLS](https://web.dev/cls/) which using the toggle button.
2. There is a weird underline under the degree symbol.

https://github.com/seamapi/react/assets/721372/57416bbb-479f-45a3-a96b-3399154fa352

src/lib/ui/thermostat/ThermostatCard.stories.tsx
xplato: The story needs to return something through all paths, right? I did this because it would complain on `null`.

xplato:
@razor-x in regard to CLS: this will be reduced when it's in a bounding container (I've updated the story to simulate its intended placement). The width of the card will no longer change based on the current unit; however, we will still, of course, see a width change within the card when the numbers change. This can be further reduced by making all temperatures non-floats, instead of just Fahrenheit. Up to you.

razor-x:
Ah ok thanks for the context! Let's make all temps non floats for now. We only support incrementing by integers anyway so I think this is an okay place to start.

PR #354, opened 2023-08-17
Add `celsius` values to fake climate settings

xplato:


razor-x (approved):
@xplato I think this is a bug in the fake. It should ensure those values are always returned even if they are not in the fake state (indicates the response type). Basically the initializer type should have the optional fields and the return type should not.

xplato:
@razor-x yeah, has to be. I'll look into it a bit!

PR #355, opened 2023-08-17
Truncate temperature in `ClimateSettingStatus` to avoid overflow

xplato:
Was hitting an overflow on temperature values like `15.3333333333334`—this calls `Math.trunc` to remove the decimal (should we round them first?).


razor-x:
Closing in favor of https://github.com/seamapi/react/pull/367 per internal discussion.

PR #361, opened 2023-08-21
Add `Tooltip` component

xplato:
An internal component • Used in the `ThermostatDeviceDetails` view. Currently uses an info icon as the trigger, but I figure we can update the component later if it can be repurposed.

---

tooltip

https://github.com/seamapi/react/assets/87919558/ebe58719-bcae-4b21-a802-f8441b31f6ec


mikewuu (approved):
nice! 👍

src/lib/ui/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,64 @@
...
+  useEffect(() => {
+    if (isOpen) {
+      document.addEventListener('click', handleClose)
+      window.addEventListener('keydown', handleEscape)
+    }
+
+    return () => {
+      document.removeEventListener('click', handleClose)
+      window.removeEventListener('keydown', handleEscape)
+    }
```
razor-x, line 32: Prefer `globalThis` over window (friendly to downstream tooling as `globalThis` is always defined.

```suggestion
    if (isOpen) {
      globalThis.document?.addEventListener('click', handleClose)
      globalThis.addEventListener?.('keydown', handleEscape)
    }

    return () => {
      globalThis.document?.removeEventListener('click', handleClose)
      globalThis.removeEventListener?.('keydown', handleEscape)
    }
```

src/lib/ui/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,64 @@
...
+  }, [isOpen])
```
razor-x, line 33:
```suggestion
  }, [isOpen, handleEscape])
```

src/lib/ui/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const handleClose = (): void => {
+    setIsOpen(false)
+  }
+
+  const handleEscape = (ev: KeyboardEvent): void => {
+    if (ev.key === 'Escape') {
+      handleClose()
+    }
+  }
```
razor-x, line 21: Since these functions are a dep of `useEffect` you may need to wrap with `useCallback`.

src/lib/ui/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,64 @@
+import { type MouseEventHandler, useEffect, useState } from 'react'
+
+import { InfoIcon } from 'lib/icons/Info.js'
+import { InfoDarkIcon } from 'lib/icons/InfoDark.js'
+
+interface TooltipProps {
+  children: React.ReactNode
```
razor-x, line 7:
```suggestion
  children: JSX.Element
```

src/lib/ui/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,64 @@
...
+    return () => {
+      document.removeEventListener('click', handleClose)
+      window.removeEventListener('keydown', handleEscape)
+    }
+  }, [isOpen])
+
+  const handleToggle: MouseEventHandler<HTMLButtonElement> = (ev): void => {
+    ev.stopPropagation()
+    setIsOpen((prev) => !prev)
+  }
```
razor-x, line 38:
```suggestion
  const handleToggle: MouseEventHandler<HTMLButtonElement> = (event): void => {
    event.stopPropagation()
    setIsOpen((prev) => !prev)
  }
```

src/lib/ui/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,64 @@
...
+          </div>
+        </button>
+      </div>
+
+      <div
+        className='seam-tooltip-popover'
+        aria-expanded={isOpen}
+        onClick={(ev) => {
+          ev.stopPropagation()
+        }}
```
razor-x, line 58:
```suggestion
        onClick={(event) => {
          event.stopPropagation()
        }}
```

razor-x (changes requested):

src/lib/ui/Tooltip/Tooltip.tsx
xplato: This will make TS complain about strings. I'll use `PropsWithChildren` from `react` instead unless there are any objections.

razor-x: Ah yea my bad. Should have recommended that instead. If you search the existing code base there should be a lot of examples on how we currently use PropsWithChildren

PR #364, opened 2023-08-22
Remaining climate setting schedule hooks

Andrewjeska:


src/lib/seam/thermostats/climate-setting-schedules/use-delete-climate-setting-schedule.ts
```diff
@@ -0,0 +1,44 @@
...
+} from '@tanstack/react-query'
+import type { ClimateSettingScheduleDeleteRequest, SeamError } from 'seamapi'
+
+import { NullSeamClientError, useSeamClient } from 'lib/seam/use-seam-client.js'
+
+export type UseDeleteClimateSettingScheduleMutationParams =
+  ClimateSettingScheduleDeleteRequest
+
+export function useDeleteClimateSettingSchedule(): UseMutationResult<
+  unknown,
```
Andrewjeska, line 14: The endpoint returns `{}`, is `unknown` the recommended return type here?

razor-x: How about via https://github.com/typescript-eslint/typescript-eslint/issues/2063#issuecomment-675156492
```suggestion
  Record<string, never>,
```

razor-x (approved):

src/lib/seam/thermostats/climate-setting-schedules/use-delete-climate-setting-schedule.ts
Andrewjeska: Just kidding the sdk returns `{ok: true}`, I was looking at connect

https://github.com/seamapi/react/pull/364/commits/6f9e86c93f0bca4ea30b137122b4cff1a723dfee

src/lib/seam/thermostats/climate-setting-schedules/use-delete-climate-setting-schedule.ts
```diff
@@ -3,16 +3,19 @@ import {
...
 import { NullSeamClientError, useSeamClient } from 'lib/seam/use-seam-client.js'
 
 export type UseDeleteClimateSettingScheduleMutationParams =
   ClimateSettingScheduleDeleteRequest
 
-export interface UseDeleteClimateSettingScheduleData {
-  ok: true
-}
+export type UseDeleteClimateSettingScheduleData =
+  ClimateSettingScheduleDeleteResponse
```
razor-x, line 18: It looks like in the other hooks we explicitly define the type here and do not use `*Response`. I believe the reasoning for this is better editor / type dx. I think we should keep this pattern consistent in the code base, but it's something we might be able to improve on with `@seamapi/types`

```suggestion
export type UseDeleteClimateSettingScheduleData = void
```

razor-x (approved):

src/lib/seam/thermostats/climate-setting-schedules/use-delete-climate-setting-schedule.ts
Andrewjeska: Sorry, I'm getting
```
/home/runner/work/react/react/src/lib/seam/thermostats/climate-setting-schedules/use-delete-climate-setting-schedule.ts
Error:   13:51  error  void is only valid as a return type or generic type argument  @typescript-eslint/no-invalid-void-type

```

Can we keep it as it was before?

Andrewjeska: I get the same thing if I pass in `void` to `UseMutationResult<>`

razor-x: Sure. This is annoying but I get it.

PR #368, opened 2023-08-23
rename to ClimateSettingScheduleRowDetails

Andrewjeska:
Name conflicted with the upcoming component `ClimateSettingScheduleDetails`


razor-x (approved):
Could do `ClimateSettingScheduleRowDetails`. It's an local sub component so the name is not so important, but I agree it's nice not to have an internal component share a name with a public one

Andrewjeska:
> Could do `ClimateSettingScheduleRowDetails`. It's an local sub component so the name is not so important, but I agree it's nice t avoid having an internal component share a name with a public one.

https://github.com/seamapi/react/pull/368/commits/28267d7d394715ce6d287244f597170337c6c7ca

PR #372, opened 2023-08-24
Add `TemperatureControl`

xplato:
Adds the `TemperatureControl` component • a modular component containing an input range and some increment and decrement buttons. I figured it best to leave this PR as uncomplicated as possible, so I'm going to add the `TemperatureControlGroup` (I might name this `ClimateControl` or `ClimateControlGroup` instead) component right after.

Screenshot 2023-08-29 at 1 20 11 PM

https://github.com/seamapi/react/assets/87919558/ff0fd273-2b4e-4d77-a803-66f1a1ebb7b4

**Bounds**

https://github.com/seamapi/react/assets/87919558/afbd1144-fb54-459e-b996-d36c8f849698

(Some light reading: https://css-tricks.com/sliding-nightmare-understanding-range-input/)


razor-x:
This post is a whole mood, like scrawlings of a mad CSS scientist 
image

razor-x:
Maybe more recent post: https://www.smashingmagazine.com/2021/12/create-custom-range-input-consistent-browsers/ that links to some kind of magic generator: https://range-input-css.netlify.app/

src/styles/_thermostat.scss
```diff
@@ -1,6 +1,200 @@
...
+
+      // etc.
+      &::-ms-fill-lower {
+        background: colors.$thermo-orange;
+      }
+    }
+
+    &[data-variant='cool'] {
+      // Thumbs
+      &::-webkit-slider-thumb {
```
xplato, line 93: We can't combine the selectors here; i.e.

```scss
&::-webkit-slider-thumb,
&::-moz-range-thumb,
&::-ms-thumb {
  // ...
}
```

However, using a mixin is a fair compromise for reduction of repetition.

src/styles/_thermostat.scss
```diff
@@ -1,6 +1,234 @@
 @use './colors';
 
-@mixin all {
+@mixin temperature-control-thumb(
```
xplato, line 3: 🐠 Do we want to add any stylistic differences between internal and exported mixins?

razor-x: Cc @mikewuu

mikewuu: As in class/mixin names? I don't have any strong opinions here. 

Seeing as customizations at the style/component level aren't officially supported, I think we can just aim for consistency.

src/lib/ui/thermostat/TemperatureControl.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+const meta: Meta<typeof TemperatureControl> = {
+  title: 'Library/TemperatureControl',
+  tags: ['autodocs'],
+  component: TemperatureControl,
+}
+
+type Story = StoryObj<typeof TemperatureControl>
+
+export const Content: Story = {
```
razor-x, line 15: I wonder if these should be separate stories? They would each get a name then and we wouldn't need any of the layout stuff. I think stories work best when they map one story to one component.

src/lib/ui/thermostat/TemperatureControl.tsx
```diff
@@ -0,0 +1,126 @@
...
+const DEFAULT_MIN = 50
+const DEFAULT_MAX = 90
+
+interface TemperatureControlProps {
+  variant: 'heat' | 'cool'
+  temperature: number
+  onChange: (temperature: number) => void
+  min?: number
+  max?: number
+  temperatureUnit?: 'fahrenheit' | 'celsius'
```
razor-x, line 15: How about prefix all or none? So always note temperature

```
  temperature: number
  onChange: (temperature: number) => void
  minTemperature?: number
  minTemperature?: number
  temperatureUnit?: 'fahrenheit' | 'celsius'
```

Or simplify since temperature is implied in context

```
  value: number
  onChange: (temperature: number) => void
  min?: number
  max?: number
  unit?: 'fahrenheit' | 'celsius'
```

src/lib/ui/thermostat/TemperatureControl.tsx
```diff
@@ -0,0 +1,126 @@
...
+      wrap.style.setProperty(
+        '--temperature-min',
+        input.min === '' ? String(DEFAULT_MIN) : input.min
+      )
+      wrap.style.setProperty(
+        '--temperature-max',
+        input.max === '' ? String(DEFAULT_MAX) : input.max
+      )
+    }
+  }, [temperature])
```
razor-x, line 91:
```suggestion
    if (wrapRef.current == null || inputRef.current == null) return
    const wrap = wrapRef.current
    const input = inputRef.current

    wrap.style.setProperty('--temperature-current', input.value)
    wrap.style.setProperty(
      '--temperature-min',
      input.min === '' ? String(DEFAULT_MIN) : input.min
    )
    wrap.style.setProperty(
      '--temperature-max',
      input.max === '' ? String(DEFAULT_MAX) : input.max
  }, [temperature])
```

src/lib/ui/thermostat/TemperatureControl.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const handleRangeChange: ChangeEventHandler<HTMLInputElement> = (
+    event
+  ): void => {
+    const wrap = wrapRef.current
+
+    if (wrap != null) {
+      const temperature = event.target.value
+      wrap.style.setProperty('--temperature-current', temperature)
+      onChange(Number(temperature))
+    }
```
razor-x, line 102:
```suggestion
    if (wrap == null) return
    const temperature = event.target.value
    wrap.style.setProperty('--temperature-current', temperature)
    onChange(Number(temperature))
```

src/lib/ui/thermostat/TemperatureControl.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  useEffect(() => {
+    if (wrapRef.current != null && inputRef.current != null) {
+      const wrap = wrapRef.current
+      const input = inputRef.current
+
+      wrap.style.setProperty('--temperature-current', input.value)
+      wrap.style.setProperty(
+        '--temperature-min',
+        input.min === '' ? String(DEFAULT_MIN) : input.min
```
razor-x, line 84: When would `input.min` or `input.max` be empty string?

src/lib/ui/thermostat/TemperatureControl.tsx
```diff
@@ -0,0 +1,126 @@
...
+        value={temperature}
+        onChange={handleRangeChange}
+        className='seam-temperature-range'
+        data-variant={variant}
+      />
+
+      <div className='seam-floating-temperature'>
+        <span className='seam-floating-temperature-value'>{temperature}</span>
+        <span className='seam-floating-temperature-unit'>
+          º{temperatureUnit[0]?.toUpperCase()}
```
razor-x, line 121: Clever, but can we do this explicitly instead as here so we have one kind of logic for this? https://github.com/seamapi/react/blob/1c2b2d40197a2c8dce469fad9f666f80fa1335d1/src/lib/ui/thermostat/ClimateSettingStatus.tsx#L70

razor-x:
This looks really nice! I didn't know this was possible 🤯

src/lib/ui/thermostat/TemperatureControl.tsx
xplato: Ahh, nice catch—old logic

src/lib/ui/thermostat/TemperatureControl.stories.tsx
xplato: Is this largely in contrast to the current convention? Taking a look at the `Button.stories.tsx`, `ClimateSettingStatus.stories.tsx` files, etc..

Happy to add additional stories, it would just be different than the current approach for atomic components. I kind of like the current approach where larger components have different stories for different context, but atomic components use one story for minor variations in props.

razor-x: Ok. Main feedback was when I looked at this story I didn't understand what I was looking at (I first I expected the sliders to somehow interact). [This story](https://react.seam.co/?path=/docs/library-climatesettingstatus--docs) doesn't have that issue for the most part maybe because there is no interaction.

Headings might help here. Not a big deal either way for the ui stories. They should be useful for dev purposes, so really whatever makes sense for us. One consideration is breaking up the stories will isolate the ui regression tests, so if there is an issue w/ one slider it will be clear which variant changed since the story will have a name.

src/lib/ui/thermostat/TemperatureControl.tsx
```diff
@@ -0,0 +1,124 @@
...
+  max?: number
+  unit?: 'fahrenheit' | 'celsius'
+}
+
+export function TemperatureControl({
+  variant,
+  value,
+  onChange,
+  min = DEFAULT_MIN,
+  max = DEFAULT_MAX,
```
razor-x, line 23: I don't think we need to shared consts anymore.

```suggestion
interface TemperatureControlProps {
  variant: 'heat' | 'cool'
  value: number
  onChange: (temperature: number) => void
  min?: number
  max?: number
  unit?: 'fahrenheit' | 'celsius'
}

export function TemperatureControl({
  variant,
  value,
  onChange,
  min = 50,
  max = 90,
```

razor-x (approved):

src/lib/ui/thermostat/TemperatureControl.stories.tsx
xplato: Ahh I see, okay. Maybe we can address this on a larger scale in another PR perhaps? I like the idea of component labels for stories—but also different stories is alright as well.

PR #377, opened 2023-08-28
Add Switch component

Andrewjeska:
https://github.com/seamapi/react/assets/11382084/552fb64c-8368-4f2a-a39b-1c6ebdb0a619


src/lib/ui/Switch.tsx
```diff
@@ -0,0 +1,38 @@
+import { useState } from 'react'
+
+interface SwitchProps {
+  defaultChecked?: boolean
+  label?: boolean
+  onChange?: (checked: boolean) => void
+}
```
razor-x, line 7: What if we make this a controlled component and let the consumer manage the state? They will need the current value anyway, so better to let them pass it in.

```suggestion
interface SwitchProps {
  checked: boolean
  onClick?: () => void
  label?: boolean
}
```

razor-x (commented):
Looks nice, can you add a story to the Library section so we can check it out?

src/lib/ui/Switch.tsx
razor-x: This suggestion might be off base if our general approach was to [mimic the MUI api though](https://mui.com/material-ui/react-checkbox/). CC @mikewuu

razor-x: This is all internal though so non-blocking and simple if we change our mind. Just good to align on this as we build out more ui lib component.

mikewuu: Controlled only makes sense IMO since this isn't a general UI lib. I would go a step further, and make `onChange` required, and accept the new state too:

```
onChange: (checked: boolean) => void
```

Andrewjeska: [62ade83](https://github.com/seamapi/react/pull/377/commits/62ade83739729d20cd582c4dadf95ef35952c179)

Andrewjeska:
> Looks nice, can you add a story to the Library section so we can check it out?

[62ade83](https://github.com/seamapi/react/pull/377/commits/62ade83739729d20cd582c4dadf95ef35952c179)

src/lib/ui/Switch.tsx
Andrewjeska: wait i want this toggle to work in the story one sec

src/lib/ui/Switch/Switch.tsx
```diff
@@ -0,0 +1,29 @@
+interface SwitchProps {
+  checked: boolean
+  onChange: (checked: boolean) => void
+  label?: boolean
+}
+
+function Switch({
+  checked,
+  onChange,
+  label = false,
```
razor-x, line 10: I missed this in the first review. This should be `disableLabel` since `label` implies the prop should be a text value. Alternatively, you can have `onLabel` and `offLabel` props, where if they are `null` they are hidden.

src/lib/ui/Switch/Switch.tsx
```diff
@@ -0,0 +1,29 @@
...
+}: SwitchProps): JSX.Element {
+  return (
+    <div className='seam-switch-container'>
+      {label && (
+        <label className='seam-switch-label'>{checked ? 'On' : 'Off'}</label>
```
razor-x, line 15: Should be moved to use `t.on` and `t.off`

src/lib/ui/Switch.tsx
Andrewjeska: [228aa0d](https://github.com/seamapi/react/pull/377/commits/228aa0df22c05f389c9d8015ea77221e9aa6bb70)

src/lib/ui/Switch/Switch.tsx
```diff
@@ -0,0 +1,29 @@
...
+      )}
+      <div
+        className={`seam-switch ${checked ? 'checked' : ''}`}
```
razor-x, line 18: See for example https://github.com/seamapi/react/blob/f1b539d44dad34390beeb3447ff2607e35d61ea2/src/lib/ui/Table/TableFilterBar/TableFilterItem.tsx#L19

More importantly, all class names must prefix with `seam-`

```suggestion
        className={classNames('seam-switch', { 'seam-switch-checked': checked })}
```

razor-x (changes requested):

src/lib/ui/Switch/Switch.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+type Story = StoryObj<typeof Switch>
+
+export const Content: Story = {
+  render: (props) => {
+    const [checked, setChecked] = useState(props.checked || false)
+
+    // allows the prop from the story to update the component
+    useEffect(() => {
+      setChecked(props.checked)
+    }, [props.checked])
```
razor-x, line 21: I tried this on the story and it almost works, but toggling the switch doesn't update the Storybook control to match. I wonder if there is a hook we can use from Storybook that will sync the state for us. 

Not a big deal for this story, but I'd love to understand the controls a bit better to enhance the Storybook devx overall.

Andrewjeska: From GPT:

> Unfortunately, Storybook doesn't provide out-of-the-box two-way data binding for its controls. The controls in Storybook's UI will not automatically update when the state within a component (or a story) changes. So while you can set initial values via args and use those values inside your components, changes to internal component state will not propagate back to the controls.
> 
> However, if you truly want a mechanism where toggling the switch updates the Storybook controls, you'd likely have to make use of story actions and some additional custom setup, but this won't be a trivial task.
> 
> One simple way to provide feedback within Storybook without changing the controls would be to use an action to log switch state changes. Actions can be used to display data in the "Actions" panel in Storybook:

I also wanted the storybook control to update but it might be tough. lemme look more into it.

razor-x: Ok, this is very much a nice to have. Mostly something I've been curious about but haven't had the opportunity to dig further myself.

Andrewjeska: Oh baby!

src/lib/ui/Switch/Switch.tsx
Andrewjeska: https://github.com/seamapi/react/pull/377/commits/ea36ce2ba9d59df08e9b2616ac49878a31cfd1b6#diff-b3283d61cb3fff4ad3dd0313780301f04779102940031eadd5fd19e0e2e1c9efR33

src/lib/ui/Switch/Switch.tsx
Andrewjeska: https://github.com/seamapi/react/pull/377/commits/ea36ce2ba9d59df08e9b2616ac49878a31cfd1b6#diff-b3283d61cb3fff4ad3dd0313780301f04779102940031eadd5fd19e0e2e1c9efR20

src/lib/ui/Switch/Switch.stories.tsx
Andrewjeska: https://sandroroth.com/blog/storybook-controlled-components

Andrewjeska: https://github.com/seamapi/react/pull/377/commits/ea36ce2ba9d59df08e9b2616ac49878a31cfd1b6#diff-74b781a75c672defb3e830dd400bb08c08c6f4b807c10957db4593d8726a8ff6R1-R30

src/lib/ui/Switch/Switch.tsx
```diff
@@ -0,0 +1,38 @@
...
+export interface SwitchProps {
+  checked: boolean
+  onChange: (checked: boolean) => void
+  disableLabel?: boolean
+}
+
+function Switch({
+  checked,
+  onChange,
+  disableLabel = true,
```
razor-x, line 12: Minor note, but default optional prop convention in this project is that missing is the same as falsy. This makes usage really clean since you can do `<Foo enableBar />` or `<Foo disableBar />` to opt in to the non-default Boolean behaivor.

So either the default should be to enable the label

```
  disableLabel = fasle,
```

or the prop name would change to match the default of opting in to show the label

```
  enableLabel = false,
```

razor-x (approved):

src/lib/ui/Switch/Switch.tsx
Andrewjeska: I'll go with enableLabel, good catch

Andrewjeska: [05a1f27](https://github.com/seamapi/react/pull/377/commits/05a1f273fc62fd6c1b94970c139883877c6951ba)

PR #378, opened 2023-08-28
ClimateSettingScheduleCard

Andrewjeska:
Screenshot 2023-08-28 at 5 10 27 PM


src/lib/ui/thermostat/ClimateSettingScheduleCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </span>
+    )
+
+  return <span>{t.expired}</span>
+}
+
+const t = {
+  starts: 'Starts',
+  ends: 'Ends',
+  expired: 'Expired',
```
Andrewjeska, line 97: I need to clarify what this should say, this is a placeholder for now

.storybook/seed-fake.js
```diff
@@ -471,6 +471,22 @@ export const seedFake = (db) => {
...
+    name: 'Guest - Robinson with a very long title to test truncation and text wrap',
+    schedule_starts_at: '2023-07-04T00:00:00.000Z',
+    schedule_ends_at: '2027-08-14T00:00:00.000Z',
+    schedule_type: 'time_bound',
+    manual_override_allowed: true,
+    automatic_heating_enabled: false,
+    automatic_cooling_enabled: true,
+    hvac_mode_setting: 'cool',
+    cooling_set_point_fahrenheit: 72,
+    cooling_set_point_celsius: 22,
```
razor-x, line 487:
```suggestion
    cooling_set_point_celsius: 22.2222,
```

src/lib/dates.ts
```diff
@@ -138,3 +138,20 @@ export const createIsoDate = (date: string, timezone: string): string => {
...
+ *
+ */
+export const formatDateAndTime = (date: string): string =>
+  `${DateTime.fromISO(date).toLocaleString({
+    month: 'short',
+    day: 'numeric',
+  })} at ${DateTime.fromISO(date).toLocaleString({
+    hour: 'numeric',
+    minute: '2-digit',
+  })}`
```
razor-x, line 157: Let's use proper locale formatting and format the full date and time together, e.g., not mix locale w/ English text ("at").

Does something like this work?
```
export const formatDateAndTime = (date: string): string =>
  DateTime.fromISO(date).toLocaleString(
    month: 'short',
    day: 'numeric',
    hour: 'numeric',
    minute: '2-digit',
  )}
```

Or maybe just 

`DateTime.fromISO(date).toLocaleString(DateTime.DATETIME_FULL)`

src/lib/ui/thermostat/ClimateSettingScheduleDevice.tsx
```diff
@@ -0,0 +1,54 @@
+import { isThermostatDevice, type ThermostatDevice } from 'seamapi'
+
+import { ChevronRightIcon } from 'lib/icons/ChevronRight.js'
+import { useDevice } from 'lib/seam/devices/use-device.js'
+import { DeviceImage } from 'lib/ui/device/DeviceImage.js'
+import { ClimateSettingStatus } from 'lib/ui/thermostat/ClimateSettingStatus.js'
+
+export function ClimateSettingDevice({
```
razor-x, line 8: What is a `ClimateSettingScheduleDevice`? Is it different than a thermostat?

src/lib/ui/thermostat/ClimateSettingScheduleDevice.tsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <div className='seam-climate-setting-schedule-device-bar'>
+      <div className='seam-climate-setting-schedule-device-content'>
+        <div className='seam-device-image'>
+          <DeviceImage device={device} />
+        </div>
+        <div className='seam-climate-setting-schedule-device-details'>
+          <div className='seam-device-name'>{device.properties.name}</div>
+          <div className='seam-device-current-climate-setting'>
+            Current climate:
```
razor-x, line 44:
```suggestion
            {t.currentClimate}:
```

src/lib/ui/thermostat/ClimateSettingScheduleDevice.tsx
```diff
@@ -0,0 +1,54 @@
...
+            <ClimateSettingStatus
+              climateSetting={device.properties.current_climate_setting}
+            />
+          </div>
+        </div>
+      </div>
+      <ChevronRightIcon className='chevron' />
```
razor-x, line 51: Cannot have any non prefixed classnames.

src/lib/ui/thermostat/ClimateSettingScheduleDevice.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useDevice } from 'lib/seam/devices/use-device.js'
+import { DeviceImage } from 'lib/ui/device/DeviceImage.js'
+import { ClimateSettingStatus } from 'lib/ui/thermostat/ClimateSettingStatus.js'
+
+export function ClimateSettingDevice({
+  deviceId,
+}: {
+  deviceId: string
+}): JSX.Element | null {
+  const { isLoading, device } = useDevice({
```
razor-x, line 13: I need to check on this but I think we will probably limit using the api hooks to the components in `lib/seam`. Instead this component should take an object with type compatible with a ThermostateDevice

src/lib/ui/thermostat/ClimateSettingScheduleCard.tsx
```diff
@@ -0,0 +1,99 @@
+import type { ClimateSettingSchedule } from 'seamapi'
+
+import { formatDateAndTime } from 'lib/dates.js'
+import { ClimateSettingScheduleIcon } from 'lib/icons/ClimateSettingSchedule.js'
+import { useClimateSettingSchedule } from 'lib/seam/thermostats/climate-setting-schedules/use-climate-setting-schedule.js'
+import { ClimateSettingDevice } from 'lib/ui/thermostat/ClimateSettingScheduleDevice.js'
+import { ClimateSettingStatus } from 'lib/ui/thermostat/ClimateSettingStatus.js'
+
+import { DotDivider } from '../layout/DotDivider.js'
```
razor-x, line 9:
```suggestion
import { DotDivider } from 'lib/ui/layout/DotDivider.js'
```

src/lib/ui/thermostat/ClimateSettingScheduleCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+    </div>
+  )
+}
+
+function Content(props: {
+  climateSettingScheduleId: string
+}): JSX.Element | null {
+  const { climateSettingScheduleId } = props
+
+  const { climateSettingSchedule } = useClimateSettingSchedule(
```
razor-x, line 30: Same note about keeping api hooks out of ui.

src/lib/ui/thermostat/ClimateSettingScheduleCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+    climateSettingScheduleId
+  )
+
+  if (climateSettingSchedule == null) {
+    return null
+  }
+
+  const name = climateSettingSchedule.name ?? t.fallbackName
```
razor-x, line 38: Interesting that we allow name to be null on the api 🤔

src/lib/ui/thermostat/ClimateSettingScheduleCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+  if (startTime <= Date.now() && Date.now() <= endTime)
+    return (
+      <span>
+        {`${t.ends} ${formatDateAndTime(
+          climateSettingSchedule.schedule_starts_at
+        )}`}
+      </span>
+    )
+
+  return <span>{t.expired}</span>
```
razor-x, line 91: Prefer luxon here for all "date stuff"

src/lib/ui/thermostat/ClimateSettingScheduleCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+  if (Date.now() < startTime)
+    return (
+      <span>
+        {`${t.starts} ${formatDateAndTime(
+          climateSettingSchedule.schedule_starts_at
+        )}`}
+      </span>
+    )
+
+  if (startTime <= Date.now() && Date.now() <= endTime)
```
razor-x, line 82: Logic that depends on the current time does not work without [this hook](https://github.com/seamapi/react/blob/main/src/lib/ui/use-interval.ts) or similar since it will only update on render which is not consistent.

src/lib/ui/thermostat/ClimateSettingScheduleCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+    return (
+      <span>
+        {`${t.ends} ${formatDateAndTime(
+          climateSettingSchedule.schedule_starts_at
+        )}`}
```
razor-x, line 87: JSX is the template 😉 
```suggestion
        {t.ends} {formatDateAndTime(climateSettingSchedule.schedule_starts_at)}
```

razor-x (changes requested):

src/lib/ui/thermostat/ClimateSettingScheduleDevice.tsx
Andrewjeska: Fair point, this was just to keep it seperate. Something with `DeviceBar` would be better.

src/lib/ui/thermostat/ClimateSettingScheduleDevice.tsx
Andrewjeska: Interesting. What's the reasoning for that? It's likely that the device is already cached at this point.

src/lib/ui/thermostat/ClimateSettingScheduleDevice.tsx
Andrewjeska: Oops, that's old.

src/lib/ui/thermostat/ClimateSettingScheduleDevice.tsx
razor-x: It's possible these components you're adding actually don't belong in UI. They look more like the sub components that will exist when we implement the top level component. Take a look at the other folders in lib/seam to see what I'm referring to.

src/lib/ui/thermostat/ClimateSettingScheduleDevice.tsx
razor-x: Might just be because of this / we haven't really implemented a component from full bottom up before https://github.com/seamapi/react/pull/378#discussion_r1308089157

src/lib/dates.ts
Andrewjeska: The figma uses`${date} at ${time}` format. I guess that part isn't critical now that I'm writing this.

Andrewjeska: Screenshot 2023-08-29 at 12 34 42 PM

This is probably fine

Andrewjeska: https://github.com/seamapi/react/pull/378/commits/dd76de64d418e71e13b2152e1c6e37b43a7c91c5#diff-be9722f143d0d492db199700788ead0bc38d9b1b4bb9401e82f0f1434d8da1ceR150-R157

src/lib/ui/thermostat/ClimateSettingScheduleDevice.tsx
Andrewjeska: https://github.com/seamapi/react/pull/378/commits/dd76de64d418e71e13b2152e1c6e37b43a7c91c5#diff-6a5322f6b608576116fc0fa481fbe912bdc8640f9ebf73bc7ea343b0216a79d9R44

src/lib/ui/thermostat/ClimateSettingScheduleDevice.tsx
Andrewjeska: https://github.com/seamapi/react/pull/378/commits/dd76de64d418e71e13b2152e1c6e37b43a7c91c5#diff-6a5322f6b608576116fc0fa481fbe912bdc8640f9ebf73bc7ea343b0216a79d9R51

src/lib/ui/thermostat/ClimateSettingScheduleCard.tsx
Andrewjeska: https://github.com/seamapi/react/pull/378/commits/dd76de64d418e71e13b2152e1c6e37b43a7c91c5#diff-be6143e4765e91974dd1ffab7a4a0c557a97a45e94ff32ca6904bbee0f72c662R67-R88

src/lib/ui/thermostat/ClimateSettingScheduleCard.tsx
Andrewjeska: https://github.com/seamapi/react/pull/378/commits/dd76de64d418e71e13b2152e1c6e37b43a7c91c5#diff-f9bff77895bd5a57c35b2cb284a3804aed48168f365c7b1ceb1f10d72756e9dd

src/lib/ui/thermostat/ClimateSettingScheduleCard.tsx
Andrewjeska: https://github.com/seamapi/react/pull/378/commits/dd76de64d418e71e13b2152e1c6e37b43a7c91c5#diff-be6143e4765e91974dd1ffab7a4a0c557a97a45e94ff32ca6904bbee0f72c662R80-R81

Andrewjeska: I often use string templating because there are instances where the space gets dropped. That could be on me not understanding when jsx puts a space in there or not.

razor-x: Ah, yea good point. The way this is currently handles is via `{' '}`, see https://github.com/seamapi/react/blob/1c2b2d40197a2c8dce469fad9f666f80fa1335d1/src/lib/seam/components/DeviceTable/DeviceRow.tsx#L60 and

razor-x: Let's pick one way to manage this issue, that way if we find a better way we can find/replace for these cases and catch them all.

Andrewjeska: Wild idea: `t.space` ?

src/lib/ui/use-current-time.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function useCurrentTime(): DateTime | null {
+  const [date, setDate] = useState<DateTime | null>(null)
+
+  const update = useCallback(() => {
+    setDate(DateTime.now())
+  }, [setDate])
+
+  useEffect(update, [update])
+  useInterval(update)
```
razor-x, line 14: Just curious, why to we need both vs just

```suggestion
  const update = useCallback(() => {
    setDate(DateTime.now())
  }, [setDate])

  useInterval(update)
```

src/lib/ui/thermostat/ClimateSettingScheduleCard.tsx
razor-x: Woa, too far man >_< But for real, in this case the space is layout not i18n related.

src/lib/dates.ts
```diff
@@ -138,3 +138,20 @@ export const createIsoDate = (date: string, timezone: string): string => {
  */
 export const getTimezoneFromIsoDate = (date: string): string | null =>
   DateTime.fromISO(date).zoneName
+
+/**
+ * Takes an ISO datetime string (2023-07-20T00:00:00.000-07:00) and returns a string of the format ${date} at ${time}
+ * (Jul 20 at 12:00 AM).
```
razor-x, line 144:
```suggestion
 * Takes an ISO datetime string (2023-07-20T00:00:00.000-07:00) and returns a string like
 * (Jul 20, 12:00 AM PDT).
```

src/lib/ui/use-current-time.ts
```diff
@@ -0,0 +1,17 @@
+import { DateTime } from 'luxon'
+import { useCallback, useEffect, useState } from 'react'
+
+import { useInterval } from 'lib/ui/use-interval.js'
+
+export function useCurrentTime(): DateTime | null {
+  const [date, setDate] = useState<DateTime | null>(null)
+
```
razor-x, line 8:
```suggestion
export function useCurrentTime(): DateTime {
  const [date, setDate] = useState<DateTime>(DateTime.now())

```

src/lib/ui/use-current-time.ts
Andrewjeska: good point

razor-x (approved):

PR #379, opened 2023-08-29
Disable `operator-no-newline-after` in Stylelint

xplato:
Disables the `scss/operator-no-newline-after` stylelint rule since this will conflict with the formatter for longer lines (which aren't very common in SCSS except for, usually, math ops).

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,6 +1,7 @@
 {
   "extends": ["stylelint-config-standard-scss"],
   "rules": {
-    "no-descending-specificity": null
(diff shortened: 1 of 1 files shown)


razor-x:
Interesting, this probably could be reported upstream somewhere since stylelint's communicated intention is not to have any conflicting prettier rules: https://stylelint.io/migration-guide/to-15/#deprecated-stylistic-rules

razor-x (approved):

xplato:
@razor-x that is interesting!

PR #381, opened 2023-08-30
Add `DetailSections`, `DetailSection`, `DetailRow`, and `AccordionRow` components

xplato:
Adds several components in the context of thermostat details—wanting to get this in so that active engs can pull from the same set of components instead of duplicating efforts.

- `DetailSections` • Used to group several `DetailSection` components together
- `DetailSection` • A group of `DetailRow` or `AccordionRow` components
- `DetailRow` • A basic detail row (like "Fan mode") that does not expand
- `AccordionRow` • A `DetailRow` that expands when clicked.

I might have went a bit overkill with the amount of stories—if so, let me know and I can consolidate.

---

In Figma, these are called "control cells" (and derivatives). I think "section" and "row" is a bit more specific, but I can rename them if anybody has any strong opinions here.


razor-x:
Are all of these meant to be used individually, or are there only one or two that will be used?

xplato:
> Are all of these meant to be used individually, or are there only one or two that will be used?

@razor-x Not sure I fully understand your question. They will all be used to construct the various detail layouts—all in different contexts. I'm not sure if there's a good way to combine them because each context is a bit different (and it'd be a little strange to, for instance, proxy the rendering through a wrapping component (like `DetailSection` having a `details` prop))

xplato:
image

Diff not found!

razor-x:
Ok, more specifically, would the Row components ever be used outside this folder?

xplato:
> Ok, more specifically, would the Row components ever be used outside this folder?

@razor-x Yeah, they will be used by the `ThermostatDeviceDetails` component under `seam/components/DeviceDetails` as well as within @Andrewjeska's components, but I'm not sure where those are located

razor-x:
> the design is slightly different
We could move this up a level to individual folders in `lib/ui/` and consider them generic ui building blocks. Components that don't use these yet can then be upgraded to use them. This will move things in a more uniform direction where any slight differences appear clearly as overrides on these components vs buried in lots of duplicated but not quite the same code / styles. Otherwise down the road, it won't be clear if slight differences are intentional or just a mistake where something copy/pasted looks like it didn't get updated.

src/lib/ui/thermostat/AccordionRow.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import { ChevronWideIcon } from 'lib/icons/ChevronWide.js'
+
+interface AccordionRowProps {
+  title: string
+  triggerRightContent?: JSX.Element
+}
+
+export function AccordionRow({
+  title,
```
razor-x, line 11: Should we avoid using `title` as a prop name since it conflict with the `title` attribute (unless it is actually passed through to some title attribute)?

We are using `heading` elsewhere to avoid this.

src/lib/ui/thermostat/AccordionRow.tsx
```diff
@@ -0,0 +1,44 @@
+import { type PropsWithChildren, useState } from 'react'
+
+import { ChevronWideIcon } from 'lib/icons/ChevronWide.js'
+
+interface AccordionRowProps {
```
razor-x, line 5: @mikewuu Introduced this and I think we should use it as it makes the (possibly exported) type match the actual props: https://github.com/seamapi/react/pull/273

I think I forgot about this and missed it in a few other PRs. It's easy to patch up again later.

```suggestion
interface AccordionRowProps extends PropsWithChildren {
```

src/lib/ui/thermostat/AccordionRow.tsx
```diff
@@ -0,0 +1,44 @@
...
+  title: string
+  triggerRightContent?: JSX.Element
```
razor-x, line 7: This reads like a verb, but it needs to be a noun. Having a hard time naming it though. Maybe

```
  expanderContent?: JSX.Element
```

Or 

```
  rightCollapsedContent?: JSX.Element
```

src/lib/ui/thermostat/AccordionRow.tsx
```diff
@@ -0,0 +1,44 @@
...
+export function AccordionRow({
+  title,
+  triggerRightContent,
+  children,
+}: PropsWithChildren<AccordionRowProps>): JSX.Element {
+  const [isExpanded, setIsExpanded] = useState(false)
+
+  const toggleExpanded = (): void => {
+    setIsExpanded((isExpanded) => !isExpanded)
+  }
```
razor-x, line 19: Can we use `useToggle` here instead?

src/lib/ui/thermostat/AccordionRow.tsx
xplato: Ahh yes, thanks for pointing this out!

xplato: Do you think `heading` is a bit suggestive here? Would `label` work or do you just prefer `heading`?

razor-x: Label works too. All the options have some corresponding  html thing, `title` has an attribute / tag, `heading` = h1,h2,..., `label` has `<label>`. The main reason we avoided `title` is actually because we need our props to never conflict with attributes since they are make into custom elements. That rule doesn't really apply to internal components. So keeping title is would be ok. We just have to pick something that maps closely to the intent and not introduce too many different names in the code for the "same thing" across our internal ui lib.

src/lib/ui/thermostat/AccordionRow.tsx
xplato: Ahh, nice catch. I'll go with the latter I think.

src/lib/ui/thermostat/AccordionRow.tsx
xplato: Yeah totally—there's so much overlap! I think `label` actually makes the most sense in this case. I'll do that.

xplato:
> We could move this up a level to individual folders in `lib/ui/` and consider them generic ui building blocks. Components that don't use these yet can then be eventually upgraded to use them. [...]

@razor-x are you thinking just the `lib/ui/layout` folder? Or did you want a special folder just for these?

razor-x:
> > We could move this up a level to individual folders in `lib/ui/` and consider them generic ui building blocks. Components that don't use these yet can then be eventually upgraded to use them. [...]
> 
> @razor-x are you thinking just the `lib/ui/layout` folder? Or did you want a special folder just for these?

That could work. As the ui folder gets bigger we may decide to re-group things a bit once better categories emerge. I am ok with flat or folder per-component. Main thing is getting this out of the thermostat namespace so they can be used elsewhere.

src/styles/_thermostat.scss
```diff
@@ -397,7 +397,152 @@
   }
 }
 
+@mixin thermostat-detail-section {
```
razor-x, line 400: Last step would be removing `thermostat` from the style namespace.

razor-x (approved):

src/styles/_thermostat.scss
xplato: Oh good catch—thanks

PR #383, opened 2023-08-30
Add useGenerateAccessCodeCode

razor-x:
- Update seamapi and fake-seam-connect
- Add useGenerateAccessCodeCode

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "@tanstack/react-query": "^4.28.0",
         "classnames": "^2.3.2",
         "luxon": "^3.3.0",
-        "seamapi": "^8.6.4",
+        "seamapi": "^8.9.0",
         "uuid": "^9.0.0"
       },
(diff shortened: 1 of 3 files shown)


xplato (approved):

xplato:
> @xplato I really wanted to name this `useGenerateCode` but in the future things like a user credential may have a code we need to generate. This is the logical name to use that follows the other hooks, but then we end up with `CodeCode`. Open to other options before merging this.

@razor-x not sure how much we like the word "pin" but that might help reduce the repetition there. For instance, `useGenerateAccessPin`, `useAccessPinGenerator`, etc...

I don't love any of them, really. I wish we could do `useGenerateAccessCode`, but that is ambiguous...

razor-x:
Yea. We can't do pin since that is not something that exist in Seam. I think we just need to lean into the "Access Code Code" unless the API renames it.

PR #387, opened 2023-08-31
Make `label` optional for `DetailSection`

Andrewjeska:
`ClimateSettingScheduleDetails` has sections without labels

Screenshot 2023-08-31 at 12 25 09 PM

### DIFF
--- a/src/lib/ui/layout/DetailSection.tsx
+++ b/src/lib/ui/layout/DetailSection.tsx
@@ -3,7 +3,7 @@ import type { PropsWithChildren } from 'react'
 import { Tooltip } from 'lib/ui/Tooltip/Tooltip.js'
 
 interface DetailSectionProps {
-  label: string
+  label?: string
   tooltipContent?: string
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/ui/layout/DetailSection.tsx
```diff
@@ -15,8 +15,8 @@ export function DetailSection({
   return (
     <div className='seam-detail-section'>
       <div className='seam-detail-label-wrap'>
-        <p className='seam-detail-label'>{label}</p>
-        {tooltipContent != null && <Tooltip>{tooltipContent}</Tooltip>}
+        {label !== undefined && <p className='seam-detail-label'>{label}</p>}
+        {tooltipContent !== undefined && <Tooltip>{tooltipContent}</Tooltip>}
```
Andrewjeska, line 19: These would be undefined if not provided, no?

razor-x: Prefer to compare to null like this unless there is a reason to differentiate between null and undefined. 
```suggestion
        {label != null && <p className='seam-detail-label'>{label}</p>}
        {tooltipContent != null && <Tooltip>{tooltipContent}</Tooltip>}
```

Andrewjeska: Screenshot 2023-08-31 at 1 09 48 PM

These pop up as undefined for me though. Does `null` also work here?

razor-x (approved):

PR #388, opened 2023-08-31
Add `Temperature` component

xplato:
Simple temperature display to reduce repetitive logic across related components

### DIFF
--- a/src/lib/ui/thermostat/ClimateSettingStatus.tsx
+++ b/src/lib/ui/thermostat/ClimateSettingStatus.tsx
@@ -121,7 +121,8 @@ function Content(props: {
 
 const formatTemperatureValue = (value: number | undefined): number => {
   if (value == null) return 0
-  return Math.trunc(value)
+  if (Number.isInteger(value)) return parseFloat(value.toFixed())
+  return parseFloat(value.toFixed(1))
 }
 
 const t = {
--- a/src/lib/ui/thermostat/ClimateSettingStatus.tsx
+++ b/src/lib/ui/thermostat/ClimateSettingStatus.tsx
@@ -7,8 +7,8 @@ import { ThermostatOffIcon } from 'lib/icons/ThermostatOff.js'
 import { Temperature } from 'lib/ui/thermostat/Temperature.js'
 
 interface SetPoint {
-  fahrenheit: number
-  celsius: number
+  fahrenheit?: number
+  celsius?: number
 }
 
 interface ClimateSettingStatusProps {
@@ -30,12 +30,12 @@ export function ClimateSettingStatus({
       <Content
         mode={climateSetting.hvac_mode_setting}
         coolingSetPoint={{
-          fahrenheit: climateSetting.cooling_set_point_fahrenheit ?? 0,
-          celsius: climateSetting.cooling_set_point_celsius ?? 0,
+          fahrenheit: climateSetting.cooling_set_point_fahrenheit,
+          celsius: climateSetting.cooling_set_point_celsius,
         }}
         heatingSetPoint={{
-          fahrenheit: climateSetting.heating_set_point_fahrenheit ?? 0,
-          celsius: climateSetting.heating_set_point_celsius ?? 0,
+          fahrenheit: climateSetting.heating_set_point_fahrenheit,
+          celsius: climateSetting.heating_set_point_celsius,
         }}
         temperatureUnit={temperatureUnit}
       />
@@ -68,16 +68,18 @@ function Content(props: {
   temperatureUnit: 'fahrenheit' | 'celsius'
 }): JSX.Element | null {
   const { mode, coolingSetPoint, heatingSetPoint, temperatureUnit } = props
-  const hasCoolingSetPoint = coolingSetPoint !== undefined
-  const hasHeatingSetPoint = heatingSetPoint !== undefined
 
-  if (mode === 'cool' && hasCoolingSetPoint)
+  if (mode === 'cool' && checkSetPoint(coolingSetPoint))
     return <Temperature {...coolingSetPoint} unit={temperatureUnit} />
 
-  if (mode === 'heat' && hasHeatingSetPoint)
+  if (mode === 'heat' && checkSetPoint(heatingSetPoint))
     return <Temperature {...heatingSetPoint} unit={temperatureUnit} />
 
-  if (mode === 'heatcool' && hasHeatingSetPoint && hasCoolingSetPoint)
+  if (
+    mode === 'heatcool' &&
+    checkSetPoint(heatingSetPoint) &&
+    checkSetPoint(coolingSetPoint)
+  )
     return (
       <span>
         <Temperature {...heatingSetPoint} unit={temperatureUnit} />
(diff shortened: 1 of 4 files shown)


src/lib/ui/thermostat/Temperature.tsx
```diff
@@ -0,0 +1,26 @@
...
+    <span>
+      {Math.trunc(temperature)}°{hideUnit != null && degree}
+    </span>
+  )
+}
+
+const t = {
+  degreeFahrenheit: 'F',
+  degreeCelsius: 'C',
+}
```
razor-x, line 26:
```suggestion
interface TemperatureProps {
  fahrenheit: number
  celsius: number
  unit: 'fahrenheit' | 'celsius'
  disableUnit?: boolean
}

export function Temperature({
  fahrenheit,
  celsius,
  unit,
  disableUnit = false,
}: TemperatureProps): JSX.Element {
  const temperature = unit === 'fahrenheit' ? fahrenheit : celsius
  const degree = unit === 'fahrenheit' ? t.degreeFahrenheit : t.degreeCelsius
  return (
    <span>
      {Math.trunc(temperature)}{!disableUnit && degree}
    </span>
  )
}

const t = {
  degreeFahrenheit: '°F',
  degreeCelsius: '°C',
}
```

src/lib/ui/thermostat/Temperature.tsx
```diff
@@ -0,0 +1,26 @@
...
+  fahrenheit,
+  celsius,
+  unit,
+  hideUnit,
+}: TemperatureProps): JSX.Element {
+  const temperature = unit === 'fahrenheit' ? fahrenheit : celsius
+  const degree = unit === 'fahrenheit' ? t.degreeFahrenheit : t.degreeCelsius
+  return (
+    <span>
+      {Math.trunc(temperature)}°{hideUnit != null && degree}
```
razor-x, line 18: We should use this function anywhere we display temperature: https://github.com/seamapi/react/blob/31b83fff25332bb620070305c67ec935a91f233c/src/lib/ui/thermostat/ClimateSettingStatus.tsx#L88

src/lib/ui/thermostat/ClimateSettingStatus.tsx
```diff
@@ -59,40 +71,59 @@ function ClimateSettingIcon(props: {
...
   const hasHeatingSetPoint = heatingSetPoint !== undefined
-  const unit =
-    temperatureUnit === 'fahrenheit' ? t.degreeFahrenheit : t.degreeCelsius
 
   if (mode === 'cool' && hasCoolingSetPoint)
-    return <span>{`${coolingSetPoint}${unit}`}</span>
+    return (
+      <Temperature
+        fahrenheit={coolingSetPoint.fahrenheit}
+        celsius={coolingSetPoint.celsius}
```
razor-x, line 86: Can simplify.
```suggestion
        {...coolingSetPoint}
```

src/lib/ui/thermostat/ClimateSettingStatus.tsx
```diff
@@ -23,16 +29,22 @@ export function ClimateSettingStatus({
...
-        heatingSetPoint={formatTemperatureValue(
-          temperatureUnit === 'fahrenheit'
-            ? climateSetting.heating_set_point_fahrenheit
-            : climateSetting.heating_set_point_celsius
-        )}
+        coolingSetPoint={{
+          fahrenheit: formatTemperatureValue(
+            climateSetting.cooling_set_point_fahrenheit
+          ),
+          celsius: formatTemperatureValue(
```
razor-x, line 36: Maybe format temp function can be moved to the Temperature component.

src/lib/ui/thermostat/ClimateSettingStatus.tsx
```diff
@@ -59,40 +71,59 @@ function ClimateSettingIcon(props: {
...
   if (mode === 'off') return <span>{t.off}</span>
 
   return null
 }
 
-const formatTemperatureValue = (value: number | undefined): string => {
-  if (value == null) return '0'
-  if (Number.isInteger(value)) return value.toFixed()
-  return value.toFixed(1)
+const formatTemperatureValue = (value: number | undefined): number => {
```
razor-x, line 122: Was there a reason to change this? Keeping one decimal place was intentional. Maybe we should only show decimal for celsius?

src/lib/ui/thermostat/Temperature.tsx
xplato: This changes the underlying behavior by removing the degree symbol. I added this prop for the Thermostat card which shows the unit itself elsewhere, but still uses the degree symbol indicating its a temperature. Removing the degree symbol would just make it a normal number, essentially (unless the degree is being rendered in another context but I think that's out of scope for this component anyway)

src/lib/ui/thermostat/ClimateSettingStatus.tsx
```diff
@@ -59,40 +71,60 @@ function ClimateSettingIcon(props: {
...
+  if (value == null) return 0
+  if (Number.isInteger(value)) return parseFloat(value.toFixed())
```
razor-x, line 124: Just curious, why is `parseFloat` needed here if this is always passed a number?

src/lib/ui/thermostat/Temperature.tsx
xplato: I'll change the prop name though 👍

src/lib/ui/thermostat/ClimateSettingStatus.tsx
xplato: `.toFixed()` returns a string

src/lib/ui/thermostat/Temperature.tsx
razor-x: When would we ever display `4°` instead of `4°C` though?

src/lib/ui/thermostat/ClimateSettingStatus.tsx
razor-x: Right, the original function returned a string though since this is used to format a number as a string.

src/lib/ui/thermostat/Temperature.tsx
xplato: Sorry, I mistakenly thought the unit wasn't shown in the `TemperatureCard`—think there's a bug in that PR. I can remove this behavior altogether unless you think there's a good reason to keep it.

src/lib/ui/thermostat/ClimateSettingStatus.tsx
xplato: Yeah, I did this before I saw your comment about moving the format func, so I'll replace this behavior.

src/lib/ui/thermostat/Temperature.tsx
razor-x: Ok, sound good. Yea I agree if we a temp is shown without a unit that sounds like a bug.

src/lib/ui/thermostat/ClimateSettingStatus.tsx
```diff
@@ -23,16 +29,14 @@ export function ClimateSettingStatus({
...
-            ? climateSetting.cooling_set_point_fahrenheit
-            : climateSetting.cooling_set_point_celsius
-        )}
-        heatingSetPoint={formatTemperatureValue(
-          temperatureUnit === 'fahrenheit'
-            ? climateSetting.heating_set_point_fahrenheit
-            : climateSetting.heating_set_point_celsius
-        )}
+        coolingSetPoint={{
+          fahrenheit: climateSetting.cooling_set_point_fahrenheit ?? 0,
```
razor-x, line 33: If the temp is null, should we render anything?

razor-x (approved):

PR #410, opened 2023-09-01
Add `sublabel` to `DetailRow`

xplato:


src/lib/ui/layout/DetailRow.tsx
```diff
@@ -2,15 +2,22 @@ import type { PropsWithChildren } from 'react'
...
   label,
+  sublabel,
   children,
 }: PropsWithChildren<DetailRowProps>): JSX.Element {
   return (
     <div className='seam-detail-row'>
-      <p className='seam-row-label'>{label}</p>
+      <div className='seam-detail-row-label-wrap'>
+        <p className='seam-row-label'>{label}</p>
+        {sublabel != undefined && (
```
xplato, line 17: Using `null` here will make the element show up in the DOM, messing up the layout of rows that don't have this property. This is also a flaw in the update to `DetailSection` with the optional `label` prop—the element is still in the DOM.

src/lib/ui/layout/DetailRow.tsx
```diff
@@ -2,15 +2,22 @@ import type { PropsWithChildren } from 'react'
...
   label,
+  sublabel,
   children,
 }: PropsWithChildren<DetailRowProps>): JSX.Element {
   return (
     <div className='seam-detail-row'>
-      <p className='seam-row-label'>{label}</p>
+      <div className='seam-detail-row-label-wrap'>
+        <p className='seam-row-label'>{label}</p>
+        {sublabel !== undefined && (
```
razor-x, line 17:
```suggestion
        {sublabel != null  && (
```

xplato: This is what I mean, yeah. If we use `null` here, the element will still render in the DOM

xplato: (Same with the recent change to `DetailSection` with the optional `label` prop)

razor-x: Should be treated the same via `==`

image

xplato: See this:

https://github.com/seamapi/react/assets/87919558/f6f79acc-1620-405a-b29c-b63262a1cd2a

xplato: Sorry it's because I have `!==` not `!=`—glossed over that, my bad

PR #411, opened 2023-09-01
Add `useConnectedAccount`

xplato:
Probably not strictly required for read-only thermo components, but will be used later.


src/lib/seam/connected-accounts/use-connected-account.ts
```diff
@@ -0,0 +1,34 @@
...
+  >({
+    enabled: client != null,
+    queryKey: ['connected_accounts', 'get', normalizedParams],
+    queryFn: async () => {
+      if (client == null) return null
+      return await client.connectedAccounts.get(normalizedParams)
+    },
+  })
+
+  return { ...rest, connected_account: data }
```
razor-x, line 33:
```suggestion
  return { ...rest, connectedAccount: data }
```

razor-x (changes requested):

src/lib/seam/connected-accounts/use-connected-account.ts
```diff
@@ -0,0 +1,34 @@
...
+} from 'seamapi'
+
+import { useSeamClient } from 'lib/seam/use-seam-client.js'
+import type { UseSeamQueryResult } from 'lib/seam/use-seam-query-result.js'
+
+export type UseConnectedAccountData = ConnectedAccount | null
+
+export function useConnectedAccount(
+  params: ConnectedAccountsGetRequest
+): UseSeamQueryResult<'connected_account', UseConnectedAccountData> {
```
razor-x, line 16: Ref https://github.com/seamapi/react/blob/main/src/lib/seam/access-codes/use-access-code.ts#L17C24-L17C34
```suggestion
): UseSeamQueryResult<'connectedAccount', UseConnectedAccountData> {
```

razor-x (approved):
We could export this one as well: https://github.com/seamapi/react/blob/main/src/lib/seam/index.ts

PR #412, opened 2023-09-01
CS Schedule should link to the corresponding device  

Andrewjeska:
I think tristan's still working on device details for thermostats, just wanted this here so we're ready for it.

### DIFF
--- a/src/lib/seam/components/ClimateSettingScheduleDetails/ClimateSettingScheduleCard.tsx
+++ b/src/lib/seam/components/ClimateSettingScheduleDetails/ClimateSettingScheduleCard.tsx
@@ -12,22 +12,28 @@ import { ClimateSettingScheduleDeviceBar } from './ClimateSettingScheduleDeviceB
 
 interface ClimateSettingScheduleCardProps {
   climateSettingScheduleId: string
+  onSelectDevice: (deviceId: string) => void
 }
 
 export function ClimateSettingScheduleCard({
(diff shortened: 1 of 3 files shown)


src/lib/seam/components/ClimateSettingScheduleDetails/ClimateSettingScheduleCard.tsx
```diff
@@ -12,22 +12,28 @@ import { ClimateSettingScheduleDeviceBar } from './ClimateSettingScheduleDeviceB
 
 interface ClimateSettingScheduleCardProps {
   climateSettingScheduleId: string
+  onSelectDevice: (deviceId: string) => void
```
razor-x, line 15: I wonder if this drilled prop should just be `onDeviceClick` everywhere instead of introducing a new name.

Andrewjeska: I was following the example set by `AccessCodeDetails`

https://github.com/seamapi/react/blob/main/src/lib/seam/components/AccessCodeDetails/AccessCodeDetails.tsx#L103

razor-x: ok, lets follow that then

razor-x (approved):

PR #419, opened 2023-09-01
fix: Show default climate setting in correct section

Andrewjeska:
- Also, we should probably display the default climate setting here, not the current one


xplato (approved):
Whoops, thank you!

src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
```diff
@@ -95,10 +95,9 @@ export function ThermostatDeviceDetails(props: {
               tooltipContent={t.defaultSettingsTooltip}
             >
               <DetailRow label={t.defaultClimate}>
```
razor-x, line 97: Should this be changed?
```suggestion
              <DetailRow label={t.currentClimate}>
```

razor-x (approved):

PR #441, opened 2023-09-07
Add TemperatureControlGroup

xplato:
Adds the `TemperatureControlGroup`

## Demonstration

(The ranges for heat and cool in the demo are different, thus you will see some "strange" behavior that is in fact expected, such as at the end when I've set the delta to `0`)

https://github.com/seamapi/react/assets/87919558/258e74c3-9307-47f1-9e90-9081220f5370

## Feature support

Due to the fact that the ranges for heat and cool can be different, and the minimum delta, there's a fair amount of code ensuring that all of this is in check. Below is my "scratchpad" area which is basically pseudocode for the underlying approach.

## Scratchpad

```
H = heat
C = cool
D = delta

H_max = max heat
H_min = min heat

C_max = max cool
C_min = min cool

Actual Max Heat: min(H_max - D, C_max - D)
Actual Min Heat: max(H_min, C_min)

Actual Max Cool: min(C_max, H_max)
Actual Min Cool: max(C_min + D, H_min + D)

---

H_max = 100
H_min = 70

C_max = 90
C_min = 50

=

Actual Max Heat = 85
Actual Max Cool = 90

Actual Min Heat = 70
Actual Min Cool = 75
```


src/lib/ui/thermostat/TemperatureControlGroup.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+  title: 'Library/TemperatureControlGroup',
+  tags: ['autodocs'],
+  component: TemperatureControlGroup,
+}
+
+type Story = StoryObj<typeof TemperatureControlGroup>
+
+export const Content: Story = {
+  render: (props) => {
+    const [_, setArgs] = useArgs()
```
razor-x, line 17: We can update all the stories to use args consistently, so out of scope for this PR, but fwiw I think the intended way to define arg defaults is via argtypes shown here https://storybook.js.org/docs/react/api/arg-types#argtypes

Otherwise the control will be out of sync with the initial value.

src/lib/ui/thermostat/TemperatureControlGroup.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+    if (mode === 'cool') {
+      const { min: minCool, max: maxCool } = getCoolBounds()
+
+      if (coolValue < minCool) {
+        onCoolValueChange(minCool)
+      } else if (coolValue > maxCool) {
+        onCoolValueChange(maxCool)
+      }
+    }
```
razor-x, line 124: These conditions should always be mutually exclusive, so the `else` part is not needed here.
```suggestion
      if (heatValue < minHeat) {
        onHeatValueChange(minHeat)
      }
      if (heatValue > maxHeat) {
        onHeatValueChange(maxHeat)
      }

      if (coolValue < minCool) {
        onCoolValueChange(minCool)
      }
      if (coolValue > maxCool) {
        onCoolValueChange(maxCool)
      }
    }

    if (mode === 'heat') {
      const { min: minHeat, max: maxHeat } = getHeatBounds()

      if (heatValue < minHeat) {
        onHeatValueChange(minHeat)
      }
      if (heatValue > maxHeat) {
        onHeatValueChange(maxHeat)
      }
    }

    if (mode === 'cool') {
      const { min: minCool, max: maxCool } = getCoolBounds()

      if (coolValue < minCool) {
        onCoolValueChange(minCool)
      } 
      if (coolValue > maxCool) {
        onCoolValueChange(maxCool)
      }
    }
```

razor-x (approved):
So cool!

razor-x:
Just an idea that might help protect us from regressions: you can consider writing a jest unit test for this one. Either directly on the component state and / or factoring out the calculation functions and unit testing those.

Alternatively, might be a good first use case to try out https://storybook.js.org/docs/react/writing-tests/test-runner (see also https://github.com/seamapi/react/discussions/38#discussioncomment-6474204)

src/lib/ui/thermostat/TemperatureControlGroup.stories.tsx
xplato: Ah nice! I was having trouble finding docs on this.

PR #445, opened 2023-09-07
Fix `ChevronDownIcon` mask ref

xplato:
Same issue as #443 • Different browsers have different policies about this, it seems.


razor-x:
Anyway we can configure figma to do this for us?

xplato:
> Anyway we can configure figma to do this for us?

@razor-x I don't think this is possible natively, but you can probably make use of one of the many SVG export plugins, such as https://www.figma.com/community/plugin/814345141907543603.

Still need to edit the existing ones though

razor-x (approved):

PR #453, opened 2023-09-11
fix: Custom element name conflict with CreateAccessCodeForm 

dfoulger:
The same name was used for both CreateAccessCodeForm and EditAccessCodeForm, causing this error when trying to use native web components:

Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': the name "seam-access-code-details" has already been used with this registry
    at yO (https://react.seam.co/v/1.55.0/dist/elements.js:8251:48)
    at https://react.seam.co/v/1.55.0/dist/elements.js:28109:15


razor-x (approved):
Thanks for contributing this fix!

PR #457, opened 2023-09-11
test: Ensure unique element names

razor-x:


Andrewjeska (approved):
This looks fine, but what do the `.element.tsx	files do for us exactly?

razor-x:
They generate the custom element definitions. See https://github.com/seamapi/react/blob/main/src/elements.ts

PR #458, opened 2023-09-12
Add telemetry client

razor-x:


src/lib/seam/use-seam-client.ts
```diff
@@ -57,9 +62,14 @@ export function useSeamClient(): {
       })
 
       if (!res.ok || res.client_session?.token == null) {
-        throw new Error('Failed to get client access token')
+        throw new Error('Failed to get client session token')
       }
 
+      telemtry.identify(userIdentifierKey, {
```
razor-x, line 68: @seveibar This is where we can identify the "user". We need to choose a unique id. Maybe it should be the workspace id?

src/lib/telemetry/hooks.ts
```diff
@@ -0,0 +1,19 @@
...
+  const { client } = useTelemetryContext()
+  return client
+}
+
+export function useTelemetryOnMount(name: string): void {
+  const telemetry = useTelemetryClient()
+  useEffect(() => {
+    telemetry.track('component_mounted', {
+      component_name: name,
+      is_custom_element: globalThis.seamCustomElementNames?.length > 0,
```
razor-x, line 16: @seveibar This is reliable unless the page has loaded the element bundle AND is using react, which should never be an expected use-case.

src/lib/seam/use-seam-client.ts
```diff
@@ -11,6 +13,7 @@ export function useSeamClient(): {
   isError: boolean
   error: unknown
 } {
+  const telemtry = useTelemetryClient()
```
seveibar, line 16:
```suggestion
  const telemetry = useTelemetryClient()
```

src/lib/seam/use-seam-client.ts
seveibar: `${workspaceId}_${userIdentifierKey}` makes the most sense to me

src/lib/telemetry/client.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export class TelemetryQueue extends Queue {}
+
+export interface TelemetryClientOptions {
+  queue: TelemetryQueue
+  endpoint?: string
+  debug?: boolean
+}
+
+export class TelemetryClient {
```
seveibar, line 11: probably makes sense to add a comment since I think this is implementing an interface for segment or something?

src/lib/telemetry/hooks.ts
seveibar: also possible/maybe easier to implement this as a component wrapper, since the signature of the wrapper could be `<T>(Element:T) => T`

seveibar: I think a wrapper makes more sense because it'll be easier to inspect props or even use a `useMeasure` system

src/lib/telemetry/client.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  load(): void {
+    if (this.#loaded) return
+    this.#loaded = true
+    void this.#queue.start()
+    this.#log('Loaded')
+  }
+
+  #push = (message: QueueMessage): void => {
+    this.#queue.push(async () => {
```
razor-x, line 58: Should inject package version/name etc here somewhere

src/lib/seam/use-seam-client.ts
razor-x: We need to update the API to return the workspace id in the client session response. For cases where we only have the token, we need to make a request to get the client session.

razor-x:
@seveibar We still need to implement the API endpoints. They should return `204 No Content` or `202 Accepted` and the request is a JSON POST with the `Payload` type.

src/lib/telemetry/hooks.ts
razor-x: I tried the wrapper but ran into some additional complexity. I think the hooks are cleaner for now and we can introduce the wrapper if it helps us send more data.

seveibar: sg

seveibar (approved):

PR #464, opened 2023-09-13
Prevent relative path imports

Andrewjeska:


.gitignore
```diff
@@ -213,7 +213,8 @@ $RECYCLE.BIN/
 .LSOverride
 
 # Icon must end with two \r
-Icon
```
Andrewjeska, line 216: Apparently this is a vim thing? 

https://stackoverflow.com/questions/5843495/what-does-the-m-character-mean-in-vim

razor-x: This line should not be changed. It's useless otherwise. You can fix this via `git checkout main .gitignore` then `git add -p` and `n` to skip the line.

.gitignore
```diff
@@ -236,7 +237,7 @@ Temporary Items
 
 # VisualStudioCode
 .vscode/*
-# !.vscode/settings.json
+!.vscode/settings.json
```
razor-x, line 240: Damn. I really didn't want to add vscode settings. This is the most compelling case I've seen though.

Mainly it's not going to help non-vscode users. So I'd like to try to be inclusive of them. Is there a file we can add that will configure the behavior of the ts language server instead? Then that will work for all contributors.

Can we leave the vscode config directory out of this PR. Instead add it to the dev container config (see for example https://github.com/microsoft/vscode-dev-containers/blob/main/containers/go/.devcontainer/devcontainer.json#L19).

razor-x (commented):
OMG amazing, thank you for finding this linting rule!

.eslintrc.json
```diff
@@ -34,7 +34,8 @@
         ]
       }
     ],
-    "simple-import-sort/exports": "error"
+    "simple-import-sort/exports": "error",
+    "import/no-relative-parent-imports": "error"
```
razor-x, line 38: Can we set this to never but add overrides for certain files? I think `.stories.tsx`, `.element.tsx` and `index.ts` are the only files that should have relative imports.

Andrewjeska: "Can we set this to never even for children?"

Was there a typo here? Children of whom?

.gitignore
Andrewjeska: took it out [ee59445](https://github.com/seamapi/react/pull/464/commits/ee594451bc510218302268b8f222656dc90a7d96)

.eslintrc.json
Andrewjeska: file specific overrides: [ef0f9c9](https://github.com/seamapi/react/pull/464/commits/ef0f9c919369aacefa82dd1ffe42c42fac5da389)

razor-x: Small misunderstanding. Safe to revert ef0f9c919369aacefa82dd1ffe42c42fac5da389 since those are all child imports (non parents). The overrides are not needed (we never want relative parent imports).

.gitignore
```diff
@@ -237,7 +236,7 @@ Temporary Items
 
 # VisualStudioCode
 .vscode/*
-!.vscode/settings.json
```
razor-x: I think we still want this line commented out now since we are adding the file.

.gitignore
Andrewjeska: dev container: [96324e1](https://github.com/seamapi/react/pull/464/commits/96324e1cc50f73f660e14be5c8663e1533dc4de7)

razor-x (approved):

PR #474, opened 2023-09-19
fix(SupportedDeviceTable): model name/number wrapping

### LINKED ISSUE #427: The model number wraps
image

The 2-line (wrapping) model number looks bad in the supported devices component

**Issues:**
- The model number should not wrap
- It should be in one line in at least 90% of the cases

**Solution**
- [x] Change the responsive breakpoint for this list so the model number number column doesn’t appear until the width of the module reaches 750px. It currently appears to break at 660.
- [x] When the model number column is present, make both the model name and the model column get proportionately wider with the width of the table.
- [x] 60/40 proportionally variable column widths for the two stretchy columns.
- [x] 60 for model name. 40 for model number
- [x] Decrease font size for model and model number to 14
- This should prevent all models numbers from wrapping except the really long ones.
- [x] I don't know how many absurdly long model names and numbers there are, but we should truncate in a way that prevents wrapping to more than 2 lines in the extreme cases. Maybe something like 50 characters for names? and 40 for model numbers?

cc @sybohy

mikewuu:
fixes #427 

### Before

https://github.com/seamapi/react/assets/11449462/bbbc610e-62af-4e94-ab23-2905c2bbf4fb

### After

https://github.com/seamapi/react/assets/11449462/245daea9-f228-4fcb-a7fe-26d9e7db9b48

### DIFF
--- a/src/lib/seam/components/SupportedDeviceTable/SupportedDeviceFilterResultRow.tsx
+++ b/src/lib/seam/components/SupportedDeviceTable/SupportedDeviceFilterResultRow.tsx
@@ -37,12 +37,14 @@ export function ModelColumn({
           alt={deviceModel.brand}
           className='seam-brand-image'
         />{' '}
-        {deviceModel.model_name}
+        <div className='seam-truncated-text'>{deviceModel.model_name}</div>
       </div>
       <div className='seam-model-id'>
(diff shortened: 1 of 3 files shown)


razor-x:
@miken-seam can you take a look?

razor-x (approved):

PR #475, opened 2023-09-20
fix(AccessCodeForm): add code length validation

### LINKED ISSUE #436: Validate supported_code_lengths
When creating an access code, validate that the code input length exists in `supported_code_lengths`

mikewuu:
closes #436

### Notable Things

- Uses `react-hook-form` for validation

### Screenshots

CleanShot 2023-09-20 at 17 54 38

https://github.com/seamapi/react/assets/11449462/9c4c525a-36af-4ac8-a413-75c769ec57d1

### DIFF
--- a/src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
+++ b/src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
@@ -340,14 +340,13 @@ function getCodeLengthRequirement(device: CommonDevice): string | null {
   return codeLengths.join(', ')
 }
 
-function isSequential(numbers: string): boolean {
-  // 0 - 99 in a string
-  // 0123456789101112...99
-  const sequentialNumbers = Array.from(
-    { length: 100 },
-    (_, index) => index
-  ).join('')
+// 0 - 99 in a string
+// 0123456789101112...99
+const sequentialNumbers = Array.from({ length: 100 }, (_, index) => index).join(
+  ''
+)
 
+function isSequential(numbers: string): boolean {
   return sequentialNumbers.includes(numbers)
 }
 
--- a/src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
+++ b/src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
@@ -1,6 +1,7 @@
 import classNames from 'classnames'
 import { useState } from 'react'
-import type { AccessCode } from 'seamapi'
+import { useForm } from 'react-hook-form'
+import { type AccessCode, isLockDevice } from 'seamapi'
 
 import {
   get24HoursLater,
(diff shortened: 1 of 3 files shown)


razor-x (approved):
Code looks good, but some design questions cc @miken-seam 

1. Can we leave space for one line below the input so when the error text appears it doesn't shift the layout?
2. I wish as a user there was a way to know the code lengths before submit, like in the input label, e.g. (2-4 digits)
3. Similar question for name, like a maybe character counter like this
58716469-1826b100-8397-11e9-8c77-a73ab7055b81

razor-x:
There is something in the design to communicate field requirements before it errors
image

mikewuu:
> I wish as a user there was a way to know the code lengths before submit, like in the input label, e.g. (2-4 digits)

Do we know if code lengths _always_ sequential? ie. `3,4,5,6` can just be `3-6`.

Or should the helper text be '3,4,5,6 digit code'

mikewuu:
> 1. Can we leave space for one line below the input so when the error text appears it doesn't shift the layout?
> 3. Similar question for name, like a maybe character counter like this

Should we spin these off into a new issue, and merge this PR?

razor-x:
@mikewuu I think it's always sequential in practice, but this is not something we can assume. We can detect it though and render differently, e.g., `ifSequential(x) ? `min(x)-max(x)` : `x.join(', ')`

mikewuu:
@razor-x, added back the code validation hints in https://github.com/seamapi/react/pull/475/commits/bdf9eeb9e00872514967e917e82ba4770ba36532

- Only showing hints if `supported_code_lengths` is defined.
- If `supported_code_lengths` is defined, assume it's numbers only.

image

mikewuu:
@razor-x  good to merge this one?

src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
```diff
@@ -296,6 +319,38 @@ function getAccessCodeTimezone(
...
+    return `${codeLengths[0]}-${codeLengths[codeLengths.length - 1]}`
+  }
+
+  return codeLengths.join(', ')
+}
+
+function isSequential(numbers: string): boolean {
+  // 0 - 99 in a string
+  // 0123456789101112...99
+  const sequentialNumbers = Array.from(
```
razor-x, line 346: Probably not an real performance issue at this size, but better perhaps to initialize this array outside the function so it's only created once and not per-render.

razor-x (approved):

src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
mikewuu: Updated in https://github.com/seamapi/react/pull/475/commits/e4267e8b1b0a92269f262a778e337e7a273e60f5

PR #484, opened 2023-09-22
fix(AccessCodeForm): show generic error on failing to save code

### LINKED ISSUE #478: Show error if access code form submit fails
See design here for what the error looks like: https://github.com/seamapi/react/issues/402

We can match known errors and display the `error.message` but we should always display a generic message for unknown error codes.

mikewuu:
closes #478 

### Dependencies

- https://github.com/seamapi/react/pull/475 has updates to validation which would conflict with changes here. I've left `codeError` intact for now, but it should be cleaned up once #475  is merged.

### Screenshots

CleanShot 2023-09-22 at 12 21 49

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ export function App() {
   <seam-device-table publishable-key="your_publishable_key"></seam-device-table>
   <script
     type="module"
-    src="https://react.seam.co/v/1.58.1/dist/elements.js"
+    src="https://react.seam.co/v/1.59.0/dist/elements.js"
   ></script>
 </body>
 ```
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ export function App() {
   <seam-device-table publishable-key="your_publishable_key"></seam-device-table>
   <script
     type="module"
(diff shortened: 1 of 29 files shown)


razor-x:
@mikewuu Anyway we can test this? Maybe a ui story for AccessCodeForm?

Does the layout shift when there is an error? I'm wondering if it makes more sense to render the error below the buttons.

mikewuu:
>  Anyway we can test this? Maybe a ui story for AccessCodeForm?

@razor-x  sure, added in https://github.com/seamapi/react/pull/484/commits/6d4a84a064cddded125c8a4a0dde9e5bcca443ec


https://github.com/seamapi/react/assets/11449462/81f4ee06-b8df-47a7-8dd3-a10c7f7bcdf2

> Does the layout shift when there is an error? I'm wondering if it makes more sense to render the error below the buttons.

Yes, but I believe this is consistent with the current design spec. IMO we should leave it as is until a we get a new design spec for all components for consistency.

cc: @miken-seam

razor-x (approved):
I have a few tweaks I want to run by @miken-seam on the error placements, but this is good to merge!

PR #485, opened 2023-09-22
Add `Spinner`

xplato:
Add basic spinner component as a part of the upcoming loading styles


src/lib/ui/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,9 @@
+import classNames from 'classnames'
+
+interface SpinnerProps {
+  size?: 'small' | 'medium' | 'large'
+}
```
razor-x, line 5: We should make it required
```suggestion
interface SpinnerProps {
  size: 'small' | 'medium' | 'large'
}
```

Or set a default
```ts
 Spinner({ size = 'small' } = {}: SpinnerProps)
```

razor-x (approved):

src/lib/ui/Spinner/Spinner.tsx
xplato: Ahh thanks—will add default

PR #493, opened 2023-09-26
Add `LoadingToast`

xplato:


src/lib/ui/LoadingToast/LoadingToast.tsx
xplato, line 1: @razor-x thoughts on this initial API?

To allow animation and to keep the component isolated, we need to use props instead of conditional rendering on behalf of the consumer. I don't love it, but it's probably one of the cleaner ways to add entrance/exit animations to it.

razor-x:
@xplato The story buttons to test this are so neat!

src/lib/ui/LoadingToast/LoadingToast.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+  tags: ['autodocs'],
+  argTypes: {},
+}
+
+type Story = StoryObj<typeof LoadingToast>
+
+export const Content: Story = {
+  render: () => {
+    const [isLoading, setIsLoading] = useState(true)
+    const [isDoneLoading, setIsDoneLoading] = useState(false)
```
razor-x, line 21: useArgs here?

src/lib/ui/LoadingToast/LoadingToast.tsx
```diff
@@ -0,0 +1,58 @@
...
+}: LoadingToastProps): JSX.Element {
+  const [shouldRender, setShouldRender] = useState(true)
+
+  useEffect(() => {
+    if (isLoading) {
+      setShouldRender(true)
+    } else {
+      setShouldRender(false)
+    }
+  }, [isLoading])
```
razor-x, line 28:
```suggestion
  useEffect(() => {
    setShouldRender(isLoading)
  }, [isLoading])
```

src/lib/ui/LoadingToast/LoadingToast.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  useEffect(() => {
+    let timeout: NodeJS.Timeout
```
razor-x, line 31:
```suggestion
    let timeout: ReturnType<typeof globalThis.setTimeout>
```

src/lib/ui/LoadingToast/LoadingToast.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    if (isDoneLoading) {
+      timeout = setTimeout(() => {
```
razor-x, line 34:
```suggestion
      timeout = globalThis.setTimeout(() => {
```

src/lib/ui/LoadingToast/LoadingToast.tsx
```diff
@@ -0,0 +1,58 @@
...
+  left?: number
+}
+
+export function LoadingToast({
+  isLoading,
+  isDoneLoading = false,
+  top,
+  left,
+}: LoadingToastProps): JSX.Element {
+  const [shouldRender, setShouldRender] = useState(true)
```
razor-x, line 20: Render suggests a react rendering, hidden suggests a style, if that makes sense. 

```suggestion
  const [hidden, setHidden] = useState(false)
```

src/lib/ui/LoadingToast/LoadingToast.tsx
```diff
@@ -0,0 +1,58 @@
...
+interface LoadingToastProps {
+  isLoading: boolean
+  isDoneLoading?: boolean
+  top?: number
+  left?: number
+}
+
+export function LoadingToast({
+  isLoading,
+  isDoneLoading = false,
```
razor-x, line 16: Why do we need two props? Can we just have `isLoading` and handle the transition via timeout with one useEffect?

src/lib/ui/LoadingToast/LoadingToast.tsx
```diff
@@ -0,0 +1,50 @@
...
+      timeout = globalThis.setTimeout(() => {
+        setHidden(true)
+      }, 1000)
+    } else {
+      setHidden(false)
+    }
+
+    return () => {
+      clearTimeout(timeout)
+    }
```
razor-x, line 33: Every time I see an `else` I think if it's really needed. In this case I almost believed it's unavoidable, but squinted harder shows it's not. I can't think of a time where removing the else didn't lead to a cleaner pattern or remove a potential bug.


```suggestion
    if (!isLoading) {
      const timeout = globalThis.setTimeout(() => {
        setHidden(true)
      }, 1000)
      return () => {
        globalThis.clearTimeout(timeout)
      }
    }
    setHidden(false)
```

src/lib/ui/LoadingToast/LoadingToast.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }, [isLoading])
+
+  return (
+    <div
+      className={classNames(
+        'seam-loading-toast',
+        hidden && 'seam-loading-toast-hide'
+      )}
```
razor-x, line 41:
```suggestion
      className={classNames(
        'seam-loading-toast',
        { 'seam-loading-toast-hide': hidden }
      )}
```

src/lib/ui/LoadingToast/LoadingToast.tsx
```diff
@@ -0,0 +1,50 @@
...
+      style={{ top, left }}
+    >
+      <div className='seam-loading-toast-icon-wrap'>
+        {!isLoading ? <CheckBlackIcon /> : <Spinner size='small' />}
```
razor-x, line 45: 🐧 Reading positive bools is generally easier then holding negations in brain space.
```suggestion
        {isLoading ? <Spinner size='small' /> : <CheckBlackIcon />}
```

razor-x (approved):

src/lib/ui/LoadingToast/LoadingToast.tsx
xplato: [265edd3](https://github.com/seamapi/react/pull/493/commits/265edd36745850a8b4ff2f991c46450be64ef26b)

xplato: Going to merge this to unblock but we can open another PR if you don't like the no-op

PR #498, opened 2023-09-27
feat: All tables show a loading indicator

xplato:
Adds the `LoadingToast` to all tables (except for the `SupportedDeviceTable`).


src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
```diff
@@ -178,6 +179,10 @@ export function AccessCodeTable({
     )
   }
 
+  if (isError) {
```
xplato, line 182: @razor-x was there a reason this wasn't included before like in the `DeviceTable`?

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
```diff
@@ -200,6 +205,13 @@ export function AccessCodeTable({
             </IconButton>
           )}
         </div>
+        <div className='seam-table-header-loading-wrap'>
+          <LoadingToast
+            isLoading={isLoading}
+            label='Loading access codes'
```
razor-x, line 211: t.loading

razor-x (commented):
Can we add to the other tables as well and release as feat: All tables show a loading indicator

xplato:
@razor-x I added this to `ClimateSettingScheduleTable`, but for the `SupportedDeviceTable`, we might want @miken-seam to take a look. It has a loading state already, and with the animation it is a bit awkward. Could we hold off on this one for the time being and open an issue to revisit it right after?

razor-x:
Just at lunch now, but testing this in mobile in the example app I'm seeing some odd behavior. First the location is off center.

Second, should we suppress the toast when data is already loaded? Meaning if I navigate around I don't think I should see the loading toast on every render (the react-query API should be able to help with this)

xplato:
@razor-x Centered.

When you look at this on mobile, what size of device is it? Are any of these components mobilized? (Unless you just you're on-the-go, not on a mobile phone or whatever).

In regard to the rendering, are you just thinking `isLoading={isLoading && !isRefetching}`?

razor-x:
Looks like centering is ok now.

And yes, not showing the loader on refetch sounds reasonable

razor-x:
I feel like `isRefetching` should have fixed this, but testing here I still see the loader on every transition

https://react-hoay4cwu1-seamapi.vercel.app/examples/basic/


https://github.com/seamapi/react/assets/721372/2fcacaa5-ea03-4174-9ff9-60437a7f1179

xplato:
> I feel like `isRefetching` should have fixed this, but testing here I still see the loader on every transition

@razor-x is this not because every transition is removing the entire component from the DOM? If the component is completely de-rendered, wouldn't we expect a fresh reload when it renders again?

xplato:
@razor-x Ok, achieved with an additional timeout to remove the component from the DOM after the initial load.

razor-x (approved):

PR #500, opened 2023-09-28
Add `Snackbar`

xplato:


src/lib/ui/Snackbar/Snackbar.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }, [autoDismiss, onClose])
+
+  return (
+    <div className='seam-snackbar-wrap'>
+      <div
+        className={classNames('seam-snackbar', {
+          'seam-snackbar-hide': !isOpen,
+        })}
+      >
+        {SnackbarIcon(variant)}
```
razor-x, line 55: Why not just 

```suggestion
        <SnackbarIcon variant={variant} />
```

src/lib/ui/Snackbar/Snackbar.tsx
```diff
@@ -0,0 +1,84 @@
+import classNames from 'classnames'
+import { useEffect } from 'react'
+
+import { CheckGreenIcon } from 'lib/icons/CheckGreen.js'
+import { CloseWhiteIcon } from 'lib/icons/CloseWhite.js'
+import { ExclamationCircleIcon } from 'lib/icons/ExclamationCircle.js'
+
+const AUTO_DISMISS_MS = 5000
```
razor-x, line 8: How about making this a prop?

razor-x (approved):

src/lib/ui/Snackbar/Snackbar.tsx
xplato: There's a type error with this (`Type '{ variant: SnackbarVariant; }' is not assignable to type 'IntrinsicAttributes & SnackbarVariant'.`) and it's not consistent with what was done here: https://github.com/seamapi/react/blob/dae8377b3b083f34890bfef98f7435bfeec1a710/src/lib/ui/thermostat/ClimateModeMenu.tsx#L27

xplato: Figured with both of those, might as well leave it

razor-x: Ok to leave it, I may look into this to learn something new. Sometimes workarounds like this have unexpected gotchas if not fully understood.

PR #503, opened 2023-10-02
Menu fixes

### LINKED ISSUE #425: Filter menu usability issues
image

image

image

**Issues**:
- Sometimes the dropdown **drops up** and gets clipped by the frame above. It often works the first time, but after trying it more than once, it tends to go up instead of down (even when scrolled to the same position).
- The second menu is has other usability issues
  - Sometimes pops up. Sometimes pops down.
  - The number of brands makes it long and can require scrolling
  - There’s no apparent sort order
  - The blank item in the menu between August and Yale can be selected resulting in a broken UI
  - If the list gets longer, it will get harder to use
  - It’s not obvious how to clear the filters (“All” is easy to miss and the word “All” doesn’t reflect the task of clearing or turning off filters. 
  - It’s not clear why there is a break in the list between yale and salto.

**Solution (more work):**
- Replace with more compact UI with the second menu in a scrolling list contained in the first menu. To be added to this issue.

**Short term quicker solution:**
- Make sure menu doesn’t open upward so it won’t get clipped.
- Remove the space in the middle of the list
- Sort items alphabetically
- Make the menu shorter (<300 px tall) and add a scroll bar, so the menu doesn’t get clipped. 

@sybohy

mikewuu:
Closes #425 

See [commits](https://github.com/seamapi/react/pull/503/commits) for list of changes.

### Notable things

- Now calculating container dimensions via `body` which doesn't change depending on content.

### Demo

https://github.com/seamapi/react/assets/11449462/bce78d7b-9f65-4875-a240-f87c45d436e7

- **Filter** button is showing again.
- Opening nested menu will align to correct anchor.
- Multiple menus in storybook still open / are positioned correctly.
- Menus only open up if there is enough space. No clipping.
- Opening a menu after scrolling (and while scrolling) still positions the menu correctly.

### DIFF
--- a/src/lib/seam/components/SupportedDeviceTable/FilterCategoryMenu.tsx
+++ b/src/lib/seam/components/SupportedDeviceTable/FilterCategoryMenu.tsx
@@ -36,7 +36,10 @@ export function FilterCategoryMenu({
   onAllOptionSelect,
   buttonLabel,
 }: FilterCategoryMenuProps): JSX.Element {
-  const usableOptions = hideAllOption ? options : [allLabel, ...options]
+  const sortedOptions = [...options].sort((a, b) => a.localeCompare(b))
+  const usableOptions = hideAllOption
+    ? sortedOptions
+    : [allLabel, ...sortedOptions]
 
   return (
     <div className='seam-supported-device-table-filter-menu-wrap'>
@@ -50,32 +53,20 @@ export function FilterCategoryMenu({
         )}
       >
         <div className='seam-supported-device-table-filter-menu-content'>
-          {usableOptions
-            .sort((a, b) => {
-              // Always put 'All' first
-              if (a === allLabel) {
-                return -1
-              }
-              if (b === allLabel) {
-                return 1
-              }
-
-              return a.localeCompare(b)
-            })
-            .map((option, index) => (
-              <MenuItem
-                key={`${index}:${option}`}
-                onClick={() => {
-                  if (option === allLabel) {
-                    onAllOptionSelect?.()
-                  } else {
-                    onSelect(option)
-                  }
-                }}
-              >
-                <span>{option}</span>
-              </MenuItem>
-            ))}
+          {usableOptions.map((option, index) => (
+            <MenuItem
+              key={`${index}:${option}`}
+              onClick={() => {
+                if (option === allLabel) {
+                  onAllOptionSelect?.()
+                } else {
+                  onSelect(option)
+                }
+              }}
+            >
+              <span>{option}</span>
+            </MenuItem>
+          ))}
         </div>
       </Menu>
     </div>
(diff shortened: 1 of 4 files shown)


xplato:
@mikewuu should there be a down-chevron icon in that inner menu trigger? Looks like it's not rendering (I know this is unrelated to the changes made here). Could this be another instance of the mask ID conflicts?

src/lib/ui/Menu/Menu.tsx
```diff
@@ -97,17 +102,20 @@ export function Menu({
 
     // If the content would overflow bottom, position it above the anchor.
     const isOverFlowingBottom = bottom > containerBottom
-    const visibleTop = isOverFlowingBottom
-      ? anchorTop - contentHeight - verticalOffset
-      : top
+    const topWhenAboveAnchor = anchorTop - contentHeight - verticalOffset
+
+    // We only want to open the menu above the anchor if it won't get clipped, ie. not < 0.
```
razor-x, line 107:
```suggestion
    // Only open the menu above the anchor if it won't get clipped, i.e., not < 0.
```

src/lib/seam/components/SupportedDeviceTable/FilterCategoryMenu.tsx
```diff
@@ -49,20 +49,34 @@ export function FilterCategoryMenu({
...
-          <MenuItem
-            key={`${index}:${option}`}
-            onClick={() => {
-              if (option === allLabel) {
-                onAllOptionSelect?.()
-              } else {
-                onSelect(option)
+        <div className='seam-supported-device-table-filter-menu-content'>
+          {usableOptions
+            .sort((a, b) => {
```
razor-x, line 54: This needs to be moved above like 

```ts
const usableOptions = hideAllOption ? options : [allLabel, ...options].sort(...)
```

otherwise it will mutate the `options` array that is passed in.

src/lib/ui/Menu/Menu.tsx
```diff
@@ -42,21 +42,20 @@ export function Menu({
...
-    const el = containers[containers.length - 1]
-    if (el != null) {
-      setDocumentEl(el)
-    }
+    const documentEl = globalThis.document.documentElement
+    setDocumentEl(documentEl)
+
+    const bodyElements = documentEl?.getElementsByTagName('body')
+    if (bodyElements.length === 0) return
+    setBodyEl(bodyElements[0] as HTMLElement)
```
razor-x, line 58:
```suggestion
    const bodyElements = documentEl?.getElementsByTagName('body')
    if (bodyElements[0] == null) return
    setBodyEl(bodyElements[0])
```

mikewuu:
@xplato indeed it is. fixed in https://github.com/seamapi/react/pull/503/commits/340bb8feef9f06f443d93ad0aa3851057b5147cb

src/lib/seam/components/SupportedDeviceTable/FilterCategoryMenu.tsx
mikewuu: Thanks! I had forgotten `sort` mutated 😱 .

Updated in https://github.com/seamapi/react/pull/503/commits/912c42dcbc19ec05db4c22933d859fd5e924d864. Also allows for a much shorter `sort` fn this way.

razor-x (approved):

PR #504, opened 2023-10-02
Add `disableCloseButton` to `Snackbar`

xplato:


src/lib/ui/Snackbar/Snackbar.tsx
```diff
@@ -18,6 +18,7 @@ interface SnackbarProps {
   }
   autoDismiss?: boolean
   dismissAfterMs?: number
+  hideCloseButton?: boolean
```
razor-x, line 21: In general we have been using `disable` over `hide` for naming props that remove an element or feature from a component. One could argue well for either name, but we should stick with one convention IMO. For reference, if this prop showed the button but didn't let you click it, the name would be `closeButtonDisabled`.

```suggestion
  disableCloseButton?: boolean
```

razor-x (approved):

src/lib/ui/Snackbar/Snackbar.tsx
xplato: Ahh okay, will change

PR #505, opened 2023-10-02
Add error `Snackbar` to table components

xplato:
~~Depends on #504 and #514~~

### DIFF
--- a/.storybook/seed-fake.js
+++ b/.storybook/seed-fake.js
@@ -155,7 +155,7 @@ export const seedFake = (db) => {
     errors: [
       {
         error_code: 'failed_to_set_on_device',
-        is_device_error: true,
+        is_access_code_error: true,
         message:
           'An error occurred when we tried to set the access code on the device. We will continue to try and set the code on the device in case the error was temporary.',
       },
--- a/.storybook/seed-fake.js
+++ b/.storybook/seed-fake.js
@@ -28,7 +28,7 @@ export const seedFake = (db) => {
 
   const cw = db.addConnectWebview({
     workspace_id: ws2.workspace_id,
-    created_at: '2023-10-03T15:07:48.000',
+    created_at: '2023-05-15T15:08:49.000',
   })
 
   const ca = db.addConnectedAccount({
--- a/.storybook/seed-fake.js
+++ b/.storybook/seed-fake.js
@@ -517,5 +517,5 @@ export const seedFake = (db) => {
   })
 }
 
-export const fakePublishableKey = 'seam_pk_1'
+export const fakePublishableKey = 'seam_pk_2'
 export const fakeUserIdentifierKey = 'some_user'
--- a/.storybook/seed-fake.js
+++ b/.storybook/seed-fake.js
@@ -12,9 +12,17 @@ export const seedFake = (db) => {
     created_at: '2023-05-15T14:07:48.000',
   })
 
+  const ws3 = db.addWorkspace({
+    name: 'Seed Workspace 3 (simulated outage)',
+    publishable_key: 'seam_pk_3',
+    created_at: '2023-05-15T14:07:48.000',
+  })
+
+  db.simulateWorkspaceOutage(ws3.workspace_id)
+
   const cw = db.addConnectWebview({
     workspace_id: ws2.workspace_id,
-    created_at: '2023-05-15T15:08:49.000',
+    created_at: '2023-10-03T15:07:48.000',
   })
 
   const ca = db.addConnectedAccount({
--- a/.storybook/seed-fake.js
+++ b/.storybook/seed-fake.js
@@ -18,7 +18,13 @@ export const seedFake = (db) => {
     created_at: '2023-05-15T14:07:48.000',
   })
 
-  db.simulateWorkspaceOutage(ws3.workspace_id)
+  db.simulateWorkspaceOutage(ws3.workspace_id, {
+    routes: [
+      '/devices/list',
+      '/access_codes/list',
+      '/thermostats/climate_setting_schedules/list',
+    ],
+  })
 
   const cw = db.addConnectWebview({
     workspace_id: ws2.workspace_id,
--- a/.storybook/seed-fake.js
+++ b/.storybook/seed-fake.js
@@ -115,7 +115,7 @@ export const seedFake = (db) => {
         error_code: 'failed_to_set_on_device',
         is_access_code_error: true,
         message:
-          'An access code with the same pin already exists on the device.',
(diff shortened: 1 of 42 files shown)


src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
```diff
@@ -229,6 +226,19 @@ export function AccessCodeTable({
           disableDeleteAccessCode={disableDeleteAccessCode}
         />
       </TableBody>
+
+      {isError && (
+        <Snackbar
+          message={
+            error?.message ??
+            'An unknown error occurred fetching access codes. Please try again.'
```
xplato, line 234: Alternatives:

- `Unknown error fetching X. Please try again.`
- `Unexpected error fetching X. Please try again.`
- `An unknown error occurred fetching X. Please reload.`
- `Something went wrong. Please try again`.

razor-x: Was there a message suggested on the design? What about `loading` instead of `fetching`?

razor-x:
Any thoughts on how to trigger a loading error in a story? Maybe we need a workspace on the fake that always returns errors on get / list?

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
xplato: Updated with design recommendation

xplato:
> Any thoughts on how to trigger a loading error in a story? Maybe we need a workspace on the fake that always returns errors on get / list?

@razor-x workspace isn't a bad idea. How would you go about that?

razor-x:
We can just add a DB method to the fake to configure some behavior, like `db.simulateInternalServerError('/devices/list', { workspace_id })`

@seveibar I know you have thought about this, did you ever have a particular API in mind?

seveibar:
@razor-x that pattern seems ok, although id always suggest documenting specific known bugs or scenarios when making simulation endpoints. In this case its not a bug we're replicating in the fake and the function is unlikely to cover up a bug so i think its good!

razor-x:
That makes sense. In this case the error is a generic issue that could be caused by anything. But we could name it `simulateTemporaryOutage` following your suggestion

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
```diff
@@ -229,6 +227,22 @@ export function AccessCodeTable({
           disableDeleteAccessCode={disableDeleteAccessCode}
         />
       </TableBody>
+
+      {isError && (
+        <Snackbar
+          variant='error'
+          message={error?.message ?? 'Access codes could not be loaded'}
+          action={{
+            label: 'Try again',
```
razor-x, line 236:
```suggestion
            label: t.tryAgain,
```

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
```diff
@@ -229,6 +227,22 @@ export function AccessCodeTable({
           disableDeleteAccessCode={disableDeleteAccessCode}
         />
       </TableBody>
+
+      {isError && (
+        <Snackbar
```
razor-x, line 232: Should we build the conditional render logic into the `Snackbar` component since otherwise it it always going to be `isError && <...>`

Cleaner if we do

```
<Snackbar
  visible={isError}
```

src/lib/seam/components/ClimateSettingScheduleTable/ClimateSettingScheduleTable.tsx
```diff
@@ -163,6 +160,22 @@ export function ClimateSettingScheduleTable({
           onClimateSettingScheduleClick={handleClimateSettingScheduleClick}
         />
       </TableBody>
+
+      {isError && (
+        <Snackbar
+          variant='error'
+          message={error?.message ?? 'Climate settings could not be loaded'}
+          action={{
+            label: 'Try again',
```
razor-x, line 169: `t.` here

src/lib/seam/components/DeviceTable/DeviceTable.tsx
```diff
@@ -152,6 +149,22 @@ export function DeviceTable({
       <TableBody>
         <Content devices={filteredDevices} onDeviceClick={handleDeviceClick} />
       </TableBody>
+
+      {isError && (
+        <Snackbar
+          variant='error'
+          message={error?.message ?? 'Devices could not be loaded'}
+          action={{
+            label: 'Try again',
```
razor-x, line 158: `t.` here

razor-x (changes requested):
We need a way to test this on storybook before we can merge it to make sure it renders correctly over the components.

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
xplato: #510

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
```diff
@@ -229,6 +227,25 @@ export function AccessCodeTable({
...
+          variant='error'
+          message={
```
xplato, line 234: Should we just use `t.fallbackErrorMessage` exclusively?

xplato:
@razor-x I believe this PR should be all unblocked now. Did some local testing and your recent fix seems to have patched up all the issues I had previously.

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
```diff
@@ -268,6 +264,20 @@ export function AccessCodeTable({
             disableDeleteAccessCode={disableDeleteAccessCode}
           />
         </TableBody>
+
+        <Snackbar
+          variant='error'
+          visible={isError}
+          onClose={() => {}}
```
razor-x, line 271: Looks like onClose could be an optional prop

.storybook/seed-fake.js
```diff
@@ -28,7 +28,7 @@ export const seedFake = (db) => {
 
   const cw = db.addConnectWebview({
     workspace_id: ws2.workspace_id,
-    created_at: '2023-05-15T15:08:49.000',
+    created_at: '2023-10-03T15:07:48.000',
   })
```
razor-x, line 32: Why did we need to update this date?

razor-x (approved):

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
xplato: #526

PR #508, opened 2023-10-03
feat: Add errorFilter and warningFilter props

razor-x:
- Add errorFilter and warningFilter to common props
- Add errorFilter and warningFilter to AccessCodeHealthBar
- Simplify errorFilter

### DIFF
--- a/src/lib/seam/components/AccessCodeDetails/AccessCodeDetails.tsx
+++ b/src/lib/seam/components/AccessCodeDetails/AccessCodeDetails.tsx
@@ -6,6 +6,7 @@ import type {
   AccessCodeError,
   ConnectedAccountError,
   DeviceError,
+  SeamWarning,
 } from 'seamapi'
 
 import { useComponentTelemetry } from 'lib/telemetry/index.js'
@@ -77,7 +78,7 @@ export function AccessCodeDetails({
       variant: 'error' as const,
       message: error.message,
     })),
-    ...accessCode.warnings.map((warning) => ({
+    ...accessCode.warnings.filter(warningFilter).map((warning) => ({
       variant: 'warning' as const,
       message: warning.message,
     })),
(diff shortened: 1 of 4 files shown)


src/lib/seam/components/AccessCodeDetails/AccessCodeDetails.tsx
```diff
@@ -244,19 +245,14 @@ const formatDate = (date: string): string =>
 const errorFilter = (
   error: AccessCodeError | DeviceError | ConnectedAccountError
 ): boolean => {
-  if ('is_access_code_error' in error && !error.is_access_code_error)
```
razor-x, line 247: @xplato Do you happen to recall why the filter logic looks like it does vs this simplified version I've updated it to here?

I'm sure we had a reason, but maybe it's not relevant now or we made a mistake before?

xplato: I could be wrong, but I seem to remember that this was implemented around the time that we were reworking some errors. Maybe those error types weren't included by default like they are now? Pretty vague remembrance. Not sure exactly, unfortunately.

xplato: Pretty sure I was wrong ^

I think it's because the design includes these specific types of errors under access codes, but our API classifies these as device errors. I think this should be re-added (in the default filter), otherwise we will not see those.

xplato: Screenshot 2023-10-03 at 4 14 40 PM

^^^ That error will be hidden with the new filter logic.

razor-x: That looks like it may just be a mistake in the fake seed.

xplato: Whoops, you are right! My bad.

xplato: Will amend.

xplato: Introduced here, it seems: https://github.com/seamapi/react/commit/447563bc65d92efbc5d5000056fcbcaf60af891e

PR #510, opened 2023-10-03
Contain `Snackbar` rendering

xplato:


src/lib/ui/Snackbar/Snackbar.tsx
```diff
@@ -24,34 +23,39 @@ interface SnackbarProps {
...
   variant,
-  isOpen,
-  onClose,
+  visible,
   action,
   autoDismiss = false,
   dismissAfterMs = 5000,
   disableCloseButton = false,
 }: SnackbarProps): JSX.Element {
+  const [showSnackbar, setShowSnackbar] = useState(visible)
```
razor-x, line 32: Consider naming this hidden to match the css class and and also to avoid using a verb for a bool var

razor-x (approved):

PR #511, opened 2023-10-03
feat: Add `errorFilter` and `warningFilter` as common props

xplato:
Commandeer of #508


src/lib/filters.ts
razor-x, line 1: Can you move this to `src/lib/seam/filters.ts`

src/lib/filters.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export const deviceWarningFilter = (warning: SeamWarning): boolean => {
+  const relevantWarnings = [
+    'device_has_flaky_connection',
+    'third_party_integration_detected',
+    'salto_office_mode',
+    'salto_privacy_mode',
+    'ttlock_lock_gateway_unlocking_not_enabled',
+  ]
```
razor-x, line 30: Why was this list warning list selected? With this approach, if new warnings are released on the API it will require an update to the package.

src/lib/seam/components/common-props.tsx
```diff
@@ -1,8 +1,21 @@
...
+export type SeamCompositeError =
+  | ConnectedAccountError
+  | DeviceError
+  | AccessCodeError
 
 export interface RequiredCommonProps {
   className: string | undefined
   onBack: (() => void) | undefined
+  errorFilter: (error: SeamCompositeError) => boolean
+  warningFilter: (warning: SeamWarning) => boolean
```
razor-x, line 18: Can you move `SeamCompositeError` to filter.ts or remove it and just be non-dry for now? I'd prefer not to export this names type in the public api (this entire file is re-exported). Better error / names types will be coming to the types package and if we export this we will be introducing conflicting  / redundant types. (This is a good type, but it should not be owned by this project)

```suggestion
export interface RequiredCommonProps {
  className: string | undefined
  onBack: (() => void) | undefined
  errorFilter: (error:  ConnectedAccountError | DeviceError | AccessCodeError) => boolean
  warningFilter: (warning: SeamWarning) => boolean
  
```

razor-x:
Before merging, I just want to walk though the filter logic to ensure it will make sense from the user pov.

For example, if there is an error or warning on a table view, and they filter to a device or access code marked with the error or warning and navigate to the detail view, will they always see the error or warning, or is there a chance it gets filtered out? It would be frustrating if we created a situation where a error/warning was flagged on the table but there was no way to see the issue.

src/lib/filters.ts
xplato: Yeah, sorry, this was in poor taste—forgot to remove

src/lib/seam/components/AccessCodeTable/AccessCodeTable.tsx
```diff
@@ -273,6 +287,8 @@ function Content(props: {
         accessCodes={accessCodes}
         filter={filter}
         onFilterSelect={setFilter}
+        errorFilter={errorFilter}
```
razor-x, line 290: The main case I want to check on is if the health bar / list elements always have the same filter logic as the detail view. Same for device table. 

I'm wondering if the health bar here and whatever component renders an issue icon for a row needs to have the same filter logic as AccessCodeDetail. It's not super DRY as there is nothing ensuing the `AccessCodeDetail` has the same filter, but we should still be able to make it internally consistent.

xplato: Gotcha, okay—I will add this in now.

src/lib/seam/components/DeviceTable/DeviceHealthBar.tsx
```diff
@@ -12,22 +12,27 @@ interface DeviceHealthBarProps {
   devices: Array<UseDevicesData[number]>
   filter: AccountFilter | DeviceFilter | null
   onFilterSelect: (filter: AccountFilter | DeviceFilter | null) => void
+  errorFilter: (error: any) => boolean
```
razor-x, line 15: I think we can have a better type here.

src/lib/seam/components/DeviceTable/DeviceHealthBar.tsx
```diff
@@ -12,22 +12,27 @@ interface DeviceHealthBarProps {
...
+        .filter(errorFilter)
+        .filter((error) => 'is_connected_account_error' in error).length > 0
     )
   })
   const erroredDevices = devices.filter((device) => {
     return (
-      device.errors.filter((error) => 'is_device_error' in error).length > 0
+      device.errors
+        .filter(errorFilter)
+        .filter((error) => 'is_device_error' in error).length > 0
```
razor-x, line 35: Can we use the same functions from `filter.ts`?

razor-x (approved):

src/lib/seam/components/DeviceTable/DeviceHealthBar.tsx
xplato: Oh thank you

PR #512, opened 2023-10-03
test: Amend access code error type in fake

xplato:


.storybook/seed-fake.js
xplato, line 1: @razor-x do updates to the fake merit a release?

razor-x: no

razor-x: you can label with `test: ` or `docs: `

razor-x (approved):

Issue #462, opened 2023-09-13
Ensure query cache works with two providers


razor-x:
Ensure the react query cache will not interfere with each other if two separate sessions in two different providers are around the same page. This may require changing the client to not be a singleton or maintain a client pool per session

razor-x:
See here https://github.com/TanStack/query/discussions/3743

razor-x:
Need to enure cache is reset when client session token is changed.

PR #514, opened 2023-10-04
test: Simulate outages

xplato:
Adds the ability to simulate outages by toggling between publishable keys.

Screenshot 2023-10-03 at 6 31 13 PM

---

Note: Going from `Normal` to `Outage` sometimes requires a reload. Any ideas on how we can eliminate this?

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -46,7 +46,6 @@ const preview: Preview = {
       description: 'Simulate an outage',
       defaultValue: false,
       toolbar: {
-        icon: 'lightning',
         items: [
           { value: false, icon: 'lightning', title: 'Normal' },
           { value: true, icon: 'lightningoff', title: 'Outage' },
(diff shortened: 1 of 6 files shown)


xplato:
@razor-x I keep getting 401 errors in CI. Is this an issue on my end?

razor-x:
> Note: Going from Normal to Outage sometimes requires a reload. Any ideas on how we can eliminate this?

It may be something with the react query cache, it's possible if there is an error the behavior is to keep the latest data. Is there a way to reset the cache when switching the global arg?

Actually, this may be a more general issue I posted here: https://github.com/seamapi/react/issues/462

Might be more serious then I originally thought.

xplato:
@razor-x Yeah, probably the cache. We can manually reset it either via the `queryClient` or by a `new QueryCache()`, but I don't know the opportune time to do this (I tried adding the latter logic in `preview.tsx` but it didn't seem to have an effect; may need to be nested within the `QueryProvider`). 

Is this cache clearing logic something we'd like to add to our primary provider or do we want to encapsulate it within Storybook (if possible)?

In any case, is this a blocker for this PR?

PR #515, opened 2023-10-05
feat(AccessCodForm): show success message on save

### LINKED ISSUE #479: Implement success message after creating or editing an access code
See design here: https://github.com/seamapi/react/issues/402

Should handle https://github.com/seamapi/react/issues/491 at the same time

mikewuu:
Closes #479 

### Notable things

- adds the `notistack` package to handle snackbars. `notistack` is 24.3kb minified: https://bundlephobia.com/package/notistack@3.0.1

https://github.com/seamapi/react/assets/11449462/fe423398-d848-4626-a9ab-2e2758f7b4d1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "@tanstack/react-query": "^4.28.0",
         "classnames": "^2.3.2",
         "luxon": "^3.3.0",
+        "notistack": "^3.0.1",
         "queue": "^7.0.0",
         "react-hook-form": "^7.46.1",
         "seamapi": "^8.13.0",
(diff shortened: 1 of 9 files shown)


razor-x:
@mikewuu  We decided earlier not to use a lib for the snack bar since we are not dealing with stacking. Is there something new we need to support here that requires the 3rd party lib?

mikewuu:
ahh ok. We do still need to write a context (not local) implementation though to handle showing snackbars on route nav?

razor-x:
Maybe we can just hook into `onBack`. For consumers that use the component stand alone, they will need to use `onBack` to show their own message on the parent screen.

razor-x:
We may need to introduce `onSuccess`?

mikewuu:
Closing in favor of #517 to avoid adding new lib (stacking is not required).

PR #520, opened 2023-10-11
feat: Add onSessionUpdate prop to get client instance

razor-x:
- Run npm update
- Update all dependencies
- feat: Add onInit to SeamProvider

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,24 +96,30 @@ export function App() {
 
 #### Web component attributes and properties
 
-Each React component is defined as a custom element with a matching name in kebab-case.
-Each element is wrapped in a `SeamProvider`.
-An attribute and custom property is defined for each `SeamProvider` prop and component prop.
-Attributes are in kebab-case and properties are in snakeCase.
+Each React component is defined as a custom element
+
+- The element name is in in kebab-case,
+  e.g., `<DeviceTable>` becomes `<seam-device-table>`.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -94,6 +94,45 @@ export function App() {
 
 > Update the version in the script tag above with the exact version of this package you would like to use.
 
+#### Web component attributes and properties
+
+Each React component is defined as a custom element with a matching name in kebab-case.
```
xplato, line 99: The formatting and phrasing of this reads like a list—is it supposed to be a list?

xplato (approved):
Nice

razor-x:
@xplato Any other ideas for the prop name? I stared with `onInit` and then moved to `onSessionUpdate`. I don't hate this name but would entertain some alternatives.

xplato:
@razor-x I think both are good. Does the `client` only update once after init? You could do `onSessionInit` to make it clear that it should only change ~once (unless, of course, it's expected to update several times).

razor-x:
> @razor-x I think both are good. Does the `client` only update once after init? You could do `onSessionInit` to make it clear that it should only change ~once (unless, of course, it's expected to update several times).

@xplato No, that's why I changed the name. The method will be called whenever the client session is refreshed or changed, say if the token expires, or the user identifier key prop changes (then a new client would be created)

PR #521, opened 2023-10-11
fix(AccessCode): stale data after update

### LINKED ISSUE #491: Stale data after updating an access code
Chancing the name of an access code works, but the ui does not refresh with the new name. Maybe we can connect some update logic into this flow: https://github.com/seamapi/react/issues/479

mikewuu:
closes #491

We were previously re-fetching the AccessCode immediately after a successful update response. This would occasionally happen before the Action Attempt has been processed, resulting in stale data back.

The fix here is to just optimistically update the UI, and assume the action will eventually be successful. 

If the user refreshes, and the action hasn't been processed, they would still see the old (stale) data. Although this isn't ideal, it's still better than the current situation, which doesn't provide any feedback that the data has changed at all.

### Notable things

- Also fixed a bug where updating an `ongoing` Access Code's code wouldn't have worked.

#### Updating code from table

https://github.com/seamapi/react/assets/11449462/23191a80-9eec-4ab1-9915-2dce1013822f

#### Updating code from details

https://github.com/seamapi/react/assets/11449462/ec74b569-f3fa-4812-b035-9c81a0c8f07c

### DIFF
--- a/src/lib/seam/access-codes/use-update-access-code.ts
+++ b/src/lib/seam/access-codes/use-update-access-code.ts
@@ -64,9 +64,7 @@ export function useUpdateAccessCode(): UseMutationResult<
           }
 
           return accessCodes.map((accessCode) => {
-            const isTarget =
-              accessCode.access_code_id === variables.access_code_id
-            if (isTarget) {
+            if (accessCode.access_code_id === variables.access_code_id) {
               return {
                 ...accessCode,
                 code: variables.code ?? accessCode.code,
--- a/src/lib/seam/access-codes/use-update-access-code.ts
+++ b/src/lib/seam/access-codes/use-update-access-code.ts
@@ -4,6 +4,8 @@ import {
   useQueryClient,
 } from '@tanstack/react-query'
 import type {
+  AccessCode,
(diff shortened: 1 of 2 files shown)


src/lib/seam/access-codes/use-update-access-code.ts
```diff
@@ -35,12 +37,47 @@ export function useUpdateAccessCode(): UseMutationResult<
...
+        ['access_codes', 'list', { device_id: variables.device_id }],
+        (accessCodes) => {
+          if (accessCodes == null) {
+            return
+          }
+
+          return accessCodes.map((accessCode) => {
+            const isTarget =
+              accessCode.access_code_id === variables.access_code_id
+            if (isTarget) {
```
razor-x, line 69:
```suggestion
            if (accessCode.access_code_id === variables.access_code_id) {
```

razor-x (approved):
Nice. This should be ok since if Seam returns an action attempt then the code will update or the access code will get an error later. In fact, returning action attempts for access code update endpoint may be deprecated in the future and the api may reflect the change instead anyway.

PR #526, opened 2023-10-23
Make `onClose` prop optional in `Snackbar`

xplato:


razor-x:
Even simpler, just default onClose to ()=> {}

PR #530, opened 2023-10-23
Use fetch for api handler

razor-x:
Experimented with this PR a while ago and couldn't get it to work. Finally found the issue 🤷🏻


razor-x:
Must iterate fetch headers with `for / of`: https://github.com/node-fetch/node-fetch/issues/1006#issuecomment-1098839318

PR #525, opened 2023-10-19
feat(SupportedDeviceTable): Use updated device database 

mikewuu:
requirement from slack: https://hello-seam.slack.com/archives/C051ETBPRTQ/p1697701620130789?thread_ts=1697243012.844269&cid=C051ETBPRTQ

PR #531, opened 2023-10-24
test: Add support for Fake devicedb in tests and api

razor-x:
This is taken directly from https://github.com/seamapi/react/pull/525 where it is confirmed working. This was split off so breaking changes can be isolated to prep for v2.


api/fake-seam-connect.ts
```diff
@@ -1,23 +1,26 @@
...
-  requestStream.push(requestBuffer)
-  requestStream.push(null)
-  return requestStream
-}
+// UPSTREAM: This line must have this precise format, otherwise
+// Vercel will not include the JSON file in the deployment.
+// https://github.com/vercel/next.js/discussions/14807#discussioncomment-146735
+const devicedbSeed = readFileSync(
+  path.join(process.cwd(), '.storybook', 'devicedb-seed.json')
+)
```
seveibar, line 23: Hmm i would just import the json directly

seveibar (approved):

api/fake-seam-connect.ts
razor-x: I would need to enable `resolveJsonModule`, which normally would be ok, but for this project it will cause issues if this is used in `src` as [JSON modules](https://nodejs.org/api/esm.html#json-modules) are not usable yet.

PR #542, opened 2023-10-25
Create Climate Setting Schedule: Set default climate setting if not already set

Andrewjeska:
Screenshot 2023-10-26 at 6 31 50 PM

Still doesn't actually save yet, will follow up in another PR


src/lib/ui/ClimateSettingForm/ClimateSettingScheduleFormDefaultClimateSetting.tsx
```diff
@@ -0,0 +1,81 @@
...
+  onCancel,
+  onSave,
+}: ClimateSettingScheduleFormDefaultClimateSettingProps): JSX.Element {
+  const { device } = useDevice({
+    device_id: deviceId,
+  })
+
+  const [mode, setMode] = useState<HvacModeSetting>('heat_cool')
+  const [heatValue, setHeatValue] = useState(70)
+  const [coolValue, setCoolValue] = useState(75)
```
Andrewjeska, line 31: I don't really consider this part of the larger form, since it may oftentimes be skipped, so I'm just using state variables here. I think it's fine either way honestly.

src/lib/ui/ClimateSettingForm/ClimateSettingScheduleForm.tsx
```diff
@@ -80,13 +91,38 @@ function Content({
     | 'climate_setting'
   >('device_select')
 
+  const returnToDeviceSelect = (): void => {
+    resetField('deviceId')
+    setPage('device_select')
+  }
+
+  useEffect(() => {
+    if (page === 'device_select' && device?.properties !== undefined) {
```
razor-x, line 100:
```suggestion
    if (page === 'device_select' && device?.properties != null) {
```

src/lib/ui/ClimateSettingForm/ClimateSettingScheduleForm.tsx
```diff
@@ -80,13 +91,38 @@ function Content({
...
+      const defaultSetting = (device?.properties as ThermostatDeviceProperties)
```
razor-x, line 101: Can use `!isTheromostatDevice / return null` pattern as before.

src/lib/ui/ClimateSettingForm/ClimateSettingScheduleForm.tsx
```diff
@@ -80,13 +91,38 @@ function Content({
...
+        .default_climate_setting
+      if (defaultSetting !== undefined) setPage('name_and_schedule')
+      else setPage('default_setting')
+    }
+  }, [device?.properties, page, setPage])
+
```
Andrewjeska, line 107: From Chat over Huddle: It's probably better to do this in an onClick handler rather than using an effect

src/lib/ui/ClimateSettingForm/ClimateSettingScheduleFormDefaultClimateSetting.tsx
```diff
@@ -0,0 +1,81 @@
...
+        title={title}
+        onBack={onBack}
+        subheading={device?.properties.name}
+      />
+      <div className='seam-main'>
+        <div className='seam-climate-setting-schedule-form-default-climate-setting'>
+          <div className='seam-content'>
+            <div className='seam-default-climate-setting-message'>
+              <p>
+                This device doesn’t have a default climate set up yet. Choose
```
razor-x, line 45: `{t.}...`

razor-x (approved):

PR #543, opened 2023-10-25
Add `useUpdateThermostat`

xplato:


xplato:
@razor-x do we have any good ways of testing this outside of components?

razor-x:
This is setup: https://testing-library.com/

What do you want to test? If you want to test the hook I think it has to be tested inside a component. Maybe testing library has a pattern to get around that though

xplato:
@razor-x yeah, just testing the hook. Can this be done in a separate PR by implementing some mutation features?

src/lib/seam/thermostats/use-update-thermostat.ts
```diff
@@ -0,0 +1,91 @@
...
+  useQueryClient,
+} from '@tanstack/react-query'
+import type {
+  SeamError,
+  ThermostatDevice,
+  ThermostatsListResponse,
+  ThermostatUpdateRequest,
+} from 'seamapi'
+
+import { NullSeamClientError, useSeamClient } from 'lib/index.js'
```
razor-x, line 13:
```suggestion
import { NullSeamClientError, useSeamClient } from 'lib/seam/use-seam-client.js'
```

src/lib/seam/thermostats/use-update-thermostat.ts
```diff
@@ -0,0 +1,91 @@
...
+          })
+        }
+      )
+    },
+  })
+}
+
+function getUpdatedThermostat(
+  thermostat: ThermostatDevice,
+  variables: UseUpdateThermostatMutationParams
```
razor-x, line 70: Was this param meant to be used?

xplato: `variables`? Yes, it is used below on line 87 (unless I misunderstood)

razor-x: Ah ok, my browser search highlighter didn't work right so I missed it.

src/lib/seam/thermostats/use-update-thermostat.ts
```diff
@@ -0,0 +1,92 @@
...
+): ThermostatDevice {
+  return {
+    ...thermostat,
+
+    properties: {
+      ...thermostat.properties,
+
+      default_climate_setting: {
+        ...thermostat.properties.default_climate_setting,
+        manual_override_allowed:
```
razor-x, line 81: Can you explain this logic a bit? Why do we default to false here vs. allow the null? Should we default to `variables.<something>.manual_override_allowed`?

razor-x (approved):
Just the question on the optimistic update logic. Probably something a test would have explained.

src/lib/seam/thermostats/use-update-thermostat.ts
xplato: Yeah, I struggled with this for a bit. This is just to make TS happy. The problem is that the type for `manual_override_allowed ` when updating is `boolean | undefined`, but it has to be `boolean`—It won't allow `null`. We can also get around this by writing:

```
 default_climate_setting: {
  ...thermostat.properties.default_climate_setting,
  ...variables.default_climate_setting,
} as ClimateSetting
```

We could default to the definition in `variables`, but that would require an additional null-check and a subsequent default.

xplato: I'm going to merge this in to unblock @Andrewjeska—logic can be adjusted if needed during implementation.

razor-x: > This is just to make TS happy

I trust the types! If we type cast we may actually introduce an unexpected null.

PR #559, opened 2023-11-01
Add automatic visibility behavior to `Snackbar`

xplato:
The `automaticVisibility` prop lets the `Snackbar` component manage its own visibility, without changing existing functionality. This is useful for scenarios where visibility is determined by query results like `isError` or `isSuccess`. This feature solves the problem of needing extra logic to hide the `Snackbar`. Existing implementations are unaffected; this is an opt-in feature.

### DIFF
--- a/src/lib/ui/Snackbar/Snackbar.tsx
+++ b/src/lib/ui/Snackbar/Snackbar.tsx
@@ -1,5 +1,5 @@
 import classNames from 'classnames'
-import { useEffect } from 'react'
+import { useEffect, useState } from 'react'
 
 import { CheckGreenIcon } from 'lib/icons/CheckGreen.js'
 import { CloseWhiteIcon } from 'lib/icons/CloseWhite.js'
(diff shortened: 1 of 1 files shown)


src/lib/ui/Snackbar/Snackbar.tsx
```diff
@@ -29,29 +30,41 @@ export function Snackbar({
...
   dismissAfterMs = 5000,
   disableCloseButton = false,
+  automaticVisibility = false,
   onClose = () => {},
 }: SnackbarProps): JSX.Element {
+  const [hidden, setHidden] = useState(visible)
+
+  useEffect(() => {
+    setHidden(!visible)
+  }, [visible])
```
razor-x, line 40: Isn't this triggering a linter warning? 

```suggestion
  }, [visible, setHidden])
```

src/lib/ui/Snackbar/Snackbar.tsx
```diff
@@ -29,29 +30,41 @@ export function Snackbar({
...
+
   const { label: actionLabel, onClick: handleActionClick } = action ?? {}
 
+  const handleClose = useCallback(() => {
+    setHidden(true)
+    onClose()
+  }, [onClose])
```
razor-x, line 47: Surprised there isn't a linter error here
```suggestion
  }, [onClose, setHidden])
```

razor-x (approved):

PR #561, opened 2023-11-01
feat: Add mutation for climate settings

xplato:
Adds ability to mutate climate setting

https://github.com/seamapi/react/assets/87919558/f388f7ab-a8d6-4961-8cf4-363ac1c880a2


src/lib/debounce.ts
xplato, line 1: @razor-x I think I'd prefer to add `lodash.debounce` here instead of using this custom function—what do you think?

razor-x: What is the reason we need debounce in this case (vs have not needed it before?)

xplato: I guess it's not a requirement, but someone dragging the temperature slider could easily create 20+ queries in less than a second on both ranges. I thin debouncing would really help performance here and limit load.

Also, in the design spec, it is requested that the temperature is saved after 2 seconds of no interaction.

xplato: > vs have not needed it before?

The temperature sliders are unique UI here that would make lots of queries very quickly.


https://github.com/seamapi/react/assets/87919558/6270f88b-6bb0-4f24-b187-516c079f5c61

razor-x: Makes sense. I'd still like to avoid introducing loash. Even if we add lodash, we still need to make debounce compatible with react hooks. What about this lib: https://github.com/xnimorz/use-debounce ?

It's <1kb and is made for react.

razor-x: Semi-related, but [this lib](https://streamich.github.io/react-use/) is ok to add as a general purpose toolkit: https://github.com/seamapi/react/discussions/481

I don't think it has debounce but it does provide a lot of control flow / utils that fit into the hook patterns (vs generic FP toolkit).

razor-x: Just found this alternative to react-use which doe have debounce: https://usehooks-ts.com/react-hook/use-debounce

xplato, line 1: I think this non-hook approach is preferable for this implementation. The earlier suggested dependencies would add complexity and unnecessarily associate this debounce function with React—we don't need that for this use-case. Moreover, the final suggested solution uses `lodash` under the hood—a bit of usage by proxy.

If there are no objections, I think we should leave this as-is for this PR and perhaps revisit the topic later. I don't think it's high enough priority to block this (but please CMIIW!)

src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
```diff
@@ -234,6 +238,206 @@ function FanModeRow({ device }: { device: ThermostatDevice }): JSX.Element {
   )
 }
 
+function ClimateSettingRow({
+  device,
+}: {
+  device: ThermostatDevice
+}): JSX.Element {
+  const deviceHeatValue =
+    device.properties.current_climate_setting.heating_set_point_fahrenheit
```
xplato, line 247: At a later point, we need to add the temperature unit to some state context to feed them in to lower components. This is far beyond the scope of this PR, however.

src/lib/temperature-bounds.ts
```diff
@@ -51,3 +51,25 @@ export const getTemperatureBounds = (
   heat: getHeatBounds(controlBounds),
   cool: getCoolBounds(controlBounds),
 })
+
+export const getSupportedThermostatModes = (
+  device: ThermostatDevice
+): HvacModeSetting[] => {
+  const allModes: HvacModeSetting[] = ['heat', 'cool', 'heat_cool', 'off']
```
razor-x, line 58: Can we more this into the API?

src/lib/debounce.ts
razor-x: Ok, it's small enough. Was this function implemented from scratch, or is there a source we need to reference (stack overflow, other lib)?

src/lib/debounce.ts
```diff
@@ -0,0 +1,33 @@
+type Procedure = (...args: any[]) => void
+
+export function debounce<F extends Procedure>(
+  func: F,
+  waitMilliseconds: number
+): {
+  (this: ThisParameterType<F>, ...args: Parameters<F>): void
+  cancel: () => void
+} {
+  let timeoutId: ReturnType<typeof setTimeout> | null = null
```
razor-x, line 10:
```suggestion
  let timeoutId: ReturnType<typeof globalThis.setTimeout> | null = null
```

src/lib/temperature-bounds.ts
xplato: I think that's a good idea, but is this a blocker for this PR? I'd like to get this in before that, if possible.

src/lib/debounce.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const debouncedFunction = function (
+    this: ThisParameterType<F>,
+    ...args: Parameters<F>
+  ): void {
+    if (timeoutId !== null) {
+      clearTimeout(timeoutId)
```
razor-x, line 17:
```suggestion
      globalThis.clearTimeout(timeoutId)
```

src/lib/debounce.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+    timeoutId = setTimeout(() => {
```
razor-x, line 19:
```suggestion
    timeoutId = globalThis.setTimeout(() => {
```

src/lib/debounce.ts
```diff
@@ -0,0 +1,33 @@
...
+      timeoutId = null
+      func.apply(this, args)
+    }, waitMilliseconds)
+  }
+
+  debouncedFunction.cancel = (): void => {
+    if (timeoutId !== null) {
+      clearTimeout(timeoutId)
```
razor-x, line 27:
```suggestion
      globalThis.clearTimeout(timeoutId)
```

src/lib/debounce.ts
xplato: It was written from scratch with, IIRC, a fair amount of influence from Copilot. Any references to existing solutions are unintentional and indirect.

src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
```diff
@@ -234,6 +239,191 @@ function FanModeRow({ device }: { device: ThermostatDevice }): JSX.Element {
...
+      isSetOffSuccess
+    ) {
+      setShowSuccess(true)
+
+      const timeout = setTimeout(() => {
+        setShowSuccess(false)
+      }, 3000)
+
+      return () => {
+        clearTimeout(timeout)
```
razor-x, line 360:
```suggestion
        globalThis.clearTimeout(timeout)
```

src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
```diff
@@ -234,6 +239,191 @@ function FanModeRow({ device }: { device: ThermostatDevice }): JSX.Element {
...
+  useEffect(() => {
+    if (
+      isHeatCoolSuccess ||
+      isHeatSuccess ||
+      isCoolSuccess ||
+      isSetOffSuccess
+    ) {
+      setShowSuccess(true)
+
+      const timeout = setTimeout(() => {
```
razor-x, line 355:
```suggestion
      const timeout = globalThis.setTimeout(() => {
```

src/lib/temperature-bounds.ts
razor-x: Not a blocker.

razor-x (approved):

PR #568, opened 2023-11-08
Add `useSetThermostatOff`

xplato:


razor-x (approved):

src/lib/seam/thermostats/use-update-off.ts
```diff
@@ -0,0 +1,77 @@
...
+    },
+  })
+}
+
+function getUpdatedThermostat(thermostat: ThermostatDevice): ThermostatDevice {
+  return {
+    ...thermostat,
+    properties: {
+      ...thermostat.properties,
+      is_fan_running: false,
```
razor-x, line 74: Are we sure this always results in the fan turning off?

image

razor-x (changes requested):
Actually, quick question to resolve before merge

src/lib/seam/thermostats/use-update-off.ts
razor-x, line 1: Need to rename file.

xplato: Ooh thank you

razor-x (approved):

PR #576, opened 2023-12-14
Update dev deps

razor-x:
Needed to unblock new PRs.


mikewuu (approved):
Looks good, but why so many icon changes?

razor-x:
The icon generator dependency must have updated something minor, maybe a fix or optimization?

PR #581, opened 2024-01-08
fix: SupportedDeviceTable SeeAll button position

### LINKED ISSUE #580: Supported devices component "see all" button is in wrong place
The "see all" button should initially float over the third item in the list, but it's floating over the first item in the list. It's in the wrong place in open and closed states.

@razor-x Wasn't sure where to put this

https://github.com/seamapi/react/assets/112443289/ebb59d53-c30c-46de-90fa-6b6d2eb54c6b

image

mikewuu:
closes #580  

Should now be centered on the final row.

### Before

image

### After

CleanShot 2024-01-08 at 22 14 49

### DIFF
--- a/src/styles/_supported-device-table.scss
+++ b/src/styles/_supported-device-table.scss
@@ -173,6 +173,7 @@ $row-padding: 8px;
         0 1px 1px 0 rgb(16 27 37 / 10%),
         0 1px 8px 0 rgb(46 69 82 / 12%);
       position: absolute;
+      bottom: -13px; //  (64px row - 30px button - 8px padding) / 2
       left: 50%;
       transform: translate(-50%, -47px);
       z-index: 3;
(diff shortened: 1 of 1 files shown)


src/styles/_supported-device-table.scss
```diff
@@ -173,6 +173,7 @@ $row-padding: 8px;
         0 1px 1px 0 rgb(16 27 37 / 10%),
         0 1px 8px 0 rgb(46 69 82 / 12%);
       position: absolute;
+      bottom: -13px; //  (64px row - 30px button - 8px padding) / 2
```
razor-x, line 176: Can we use sass or css math to compute this instead of the comment?

razor-x (approved):

src/styles/_supported-device-table.scss
mikewuu: Yep, good call, thanks.

Updated in https://github.com/seamapi/react/pull/581/commits/5e67dc045b65ee18d3d4105aa67bad7973e3e515

PR #582, opened 2024-01-09
fix: remove vscode from git

mikewuu:
looks like this may have been accidentally checked in

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "typescript.preferences.importModuleSpecifier": "non-relative"
-}


razor-x:
🤷🏻 https://github.com/seamapi/react/pull/464#discussion_r1325342647

mikewuu:
@razor-x maybe no longer required, since it's in gitignore? 

https://github.com/seamapi/react/blob/da5f5e4ceaab9d1b9f47fb1d93f22db3157601d5/.gitignore#L238

cc: @Andrewjeska

razor-x:
We force added it so vscode users would not get the default import auto complete behavior. We made this exception because the linter-required import behavior is too frustrating to use with this repo with the default. Ideally vscode could detect the linter rule and match this param, but I don't think it can.

PR #583, opened 2024-01-19
Add `hideAccessCodes` flag to device table

Andrewjeska:
Screenshot 2024-01-19 at 9 48 56 PM

PR #585, opened 2024-01-22
Revert "Add `hideAccessCodes` flag to device table (#583)"

razor-x:
This reverts commit fde0c1cfe1e6cb30668c336f28a39960753252a7.

This prop introduces granular customization that is currently out of scope of this project. The recommended work around to hide arbitrary element is to target the CSS classes.


Andrewjeska:
If we're certain that the customer doesn't want this anymore, then I'm okay with it

Andrewjeska (approved):

PR #587, opened 2024-01-29
fix: Error in some contexts when using Error.captureStackTrace()

codetheweb:
This only exists in v8 contexts and was throwing in Firefox.

It's [not needed](https://stackoverflow.com/a/64063868) when extending the Error class.

### DIFF
--- a/src/lib/seam/SeamProvider.tsx
+++ b/src/lib/seam/SeamProvider.tsx
@@ -273,6 +273,5 @@ class InvalidSeamProviderProps extends Error {
   constructor(message: string) {
     super(`SeamProvider received invalid props: ${message}`)
     this.name = this.constructor.name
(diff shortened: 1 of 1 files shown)


codetheweb:
Hiding the actual error:

Screenshot 2024-01-29 at 17 42 27

dawnho (approved):

razor-x (approved):
@codetheweb Friendly reminder to use semantic release PR titles to ensure changes are published.

razor-x:
This must have been a recommended best practice at some point (maybe erroneously?): https://github.com/seamapi/javascript/blob/22bd94dfdc32a30735e5ca02b1c61fd5aae3f72c/src/lib/api-error.ts#L15-L17

We can remove it from the other libs as well.

PR #588, opened 2024-01-30
fix: Error when passing a Seam client to the SeamProvider

codetheweb:
The provider previously threw if you attempted to pass a client. Happy to add a test with some guidance, not entirely sure how to.


razor-x (approved):

razor-x:
Test is a good idea, but not blocking. Test naming / structure could mimic https://github.com/seamapi/javascript-http/blob/main/test/seam/connect/client.test.ts and https://github.com/seamapi/javascript-http/blob/main/test/seam/connect

PR #597, opened 2024-03-05
feat: Add `disableClimateSettingSchedules` prop

xplato:
## Screenshot when disabled

Screenshot 2024-03-06 at 11 19 16 AM

## Video

https://github.com/seamapi/react/assets/87919558/145ffafb-feb6-4611-be39-1d2181a5d912

### DIFF
--- a/src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
+++ b/src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
@@ -43,6 +43,7 @@ export function DeviceDetails({
     disableResourceIds,
     disableCreateAccessCode,
     disableEditAccessCode,
+    disableScheduledClimates,
     onBack,
     className,
   }
--- a/src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
+++ b/src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
@@ -37,7 +37,7 @@ export function DeviceDetails({
     return null
   }
 
-  const props = {
+  const props: Omit<DeviceDetailsProps, 'deviceId'> = {
     disableLockUnlock,
     disableDeleteAccessCode,
     disableResourceIds,
--- a/src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
+++ b/src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
@@ -23,6 +23,7 @@ export function DeviceDetails({
   disableResourceIds = false,
   disableCreateAccessCode = false,
   disableEditAccessCode = false,
+  disableScheduledClimates = false,
   onBack,
   className,
 }: DeviceDetailsProps): JSX.Element | null {
--- a/src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
+++ b/src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
@@ -23,7 +23,7 @@ export function DeviceDetails({
   disableResourceIds = false,
   disableCreateAccessCode = false,
   disableEditAccessCode = false,
-  disableScheduledClimates = false,
+  disableClimateSettingSchedules = false,
   onBack,
   className,
 }: DeviceDetailsProps): JSX.Element | null {
@@ -43,7 +43,7 @@ export function DeviceDetails({
     disableResourceIds,
     disableCreateAccessCode,
     disableEditAccessCode,
-    disableScheduledClimates,
+    disableClimateSettingSchedules,
     onBack,
     className,
   }
--- a/src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
+++ b/src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
@@ -45,7 +45,7 @@ export function ThermostatDeviceDetails({
   disableEditAccessCode,
   disableDeleteAccessCode,
   disableResourceIds = false,
-  disableScheduledClimates = false,
+  disableScheduledClimates,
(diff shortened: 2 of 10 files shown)


src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
```diff
@@ -45,6 +45,7 @@ export function ThermostatDeviceDetails({
   disableEditAccessCode,
   disableDeleteAccessCode,
   disableResourceIds = false,
+  disableScheduledClimates = false,
```
razor-x, line 48: These are internal components and should not have the default: the props should all be required. Actually the above default should be removed as well, and fixed in the lock device component too. Having these optional actually resulted in an earlier bug where the prop was not passed down.

```suggestion
  disableScheduledClimates
```

src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
```diff
@@ -36,12 +37,13 @@ export function DeviceDetails({
     return null
   }
 
-  const props = {
+  const props: Omit<DeviceDetailsProps, 'deviceId'> = {
```
razor-x, line 40: What is this needed now?

src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
```diff
@@ -36,12 +37,13 @@ export function DeviceDetails({
...
     disableLockUnlock,
     disableDeleteAccessCode,
     disableResourceIds,
     disableCreateAccessCode,
     disableEditAccessCode,
+    disableScheduledClimates,
```
razor-x, line 46:
```suggestion
    disableClimateSettingSchedules,
```

I suspect we will deprecate this  flip the default value to true in the next major version after we introduce `disableScheduledClimateSettings`.

src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
xplato: It took me a minute to figure out why the prop wasn't being passed through correctly—adding strict typing here will save others this unnecessary delay/investigation. Not required, but not harmful, I think,

src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
xplato: Ahh, good point. I've removed this default specifier on all components except for `DeviceDetails` and `DeviceTable`. Are these the two top-level ones that should remain?

src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
razor-x: Ah ok, I expanded the diff context and I see what's up 👍🏻

src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
razor-x: Yes, generally, only defaults on top level components (or UI lib components) should be allowed 👍🏻

src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
razor-x: There is no Seam API concept called "scheduled climates". Seam only provides "climate setting schedules" which will eventually be deprecated in favor of "scheduled climate settings". Thus, we should name this prop `disableClimateSettingSchedules` to match the Seam API resource. 

When we reach the deprecation stage, we will change the default value of `disableClimateSettingSchedules` to false an introduce `disableScheduledClimateSettings` instead.

src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
razor-x: Maybe we should do

```ts
interface ThermostatDeviceDetailsProps extends Required<Pick<CommonProps, '...'>> {
  device: ThermostatDevice
}

xplato: I think this is a better paradigm, yeah. It makes it clearer why global props are being used in context.

src/lib/seam/components/DeviceDetails/DeviceDetails.tsx
xplato: Gotcha—thank you for the explanation!!

src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
xplato: I think this should be done in a separate PR (right after this) for better encapsulation

xplato: Ah, I remember why there was a default for `disableResourceIds`: the linter would complain about the nullable boolean. Your type change would fix that, but would be out of the original scope. I can keep it as-is with this default now, or replace the types on this component alone here.

I think I prefer the former, then creating a separate branch to make the type specificity change on all inner-component props. What do you think?

Either is fine—just thinking out loud!

razor-x: Ok to move it out of scope of this PR.

dawnho:
Could you show a screenshot of what it looks like with this enabled?

xplato:
> Could you show a screenshot of what it looks like with this enabled?

@dawnho uploaded screenshot and video—thank you for reminding me!

razor-x (approved):

PR #598, opened 2024-03-12
Upgrade react-query to v5

phpnode:
This does change the signature of some of our hooks (`isLoading` -> `isPending`), but I'm not sure whether we consider the hooks part of our public API? cc @razor-x

### DIFF
--- a/src/lib/seam/access-codes/use-access-codes.ts
+++ b/src/lib/seam/access-codes/use-access-codes.ts
@@ -29,9 +29,7 @@ export function useAccessCodes(
     queryKey: ['access_codes', 'list', normalizedParams],
     queryFn: async () => {
       if (client == null) return []
-      return await client.accessCodes.list(normalizedParams)
-    },
-    onSuccess: (accessCodes) => {
+      const accessCodes = await client.accessCodes.list(normalizedParams)
       for (const accessCode of accessCodes) {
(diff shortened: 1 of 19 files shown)


razor-x:
Yes, this would be a new major version. 

Originally I imagined combining this with an update to the new SDK, but it's ok if we make that another major version. I'm trying not balance http://sentimentalversioning.org/ with too much major version churn.

If we want to batch some breaking changes, we can start a `next` branch instead of putting this on `main`.

src/lib/seam/access-codes/use-access-codes.ts
```diff
@@ -29,9 +29,7 @@ export function useAccessCodes(
     queryKey: ['access_codes', 'list', normalizedParams],
     queryFn: async () => {
       if (client == null) return []
-      return await client.accessCodes.list(normalizedParams)
-    },
-    onSuccess: (accessCodes) => {
+      const accessCodes = await client.accessCodes.list(normalizedParams)
       for (const accessCode of accessCodes) {
```
razor-x, line 33: Why do remove `onSuccess` here but are able to keep using it in the other hooks?

phpnode: `onSuccess` has been removed from queries but not from mutations

PR #607, opened 2024-03-15
feat: Add `disableDeviceDetails`

xplato:
Adds `disableDeviceDetails` prop to hide details in `ThermostatDeviceDetails`.

Should this be renamed to be more specific? `disableThermostatDeviceDetails`?

https://github.com/seamapi/react/assets/87919558/22b130f5-5aa9-444c-94e8-8ee58dd84f9c

### DIFF
--- a/src/lib/element.tsx
+++ b/src/lib/element.tsx
@@ -47,6 +47,7 @@ const commonProps: R2wcProps<CommonProps> = {
   disableDeleteAccessCode: 'boolean',
   disableResourceIds: 'boolean',
   disableClimateSettingSchedules: 'boolean',
+  disableDeviceDetails: 'boolean',
   onBack: 'object',
   className: 'string',
 }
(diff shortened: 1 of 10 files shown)


razor-x (changes requested):
Agree the name is confusing, it sounds like this will disable the `DeviceDetails` component somehow. I'm not sure this prop is is scope of this project. Similar to what happened with https://github.com/seamapi/react/pull/585

xplato:
@razor-x Good point. Any alternative ideas? `disableThermostatMetaDetails`? I don't like any of these 😅

razor-x:
Can we just recommend the CSS solution for this case for now?

Do you happen to know the use case or motivation for hiding just this section?

razor-x:
Closing in favor of https://github.com/seamapi/react/pull/623

PR #623, opened 2024-04-27
Add disableConnectedAccountInformation

razor-x:
- **feat: Add disableConnectedAccountInformation**
- **Move DeviceInfo to component**
- **feat: Add device info to lock DeviceDetails**


xplato (approved):
Nice, LGTM!

src/lib/seam/components/DeviceDetails/DeviceInfo.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const { connectedAccount } = useConnectedAccount(device.connected_account_id)
+  return (
+    <DetailSection label={t.deviceInfo}>
+      <DetailRow label={t.manufacturer}>
+        <div className='seam-detail-row-hstack'>
+          {device.properties.model.manufacturer_display_name}
+          {device.properties.manufacturer === 'ecobee' && <BeeIcon />}
+        </div>
+      </DetailRow>
+      {!disableConnectedAccountInformation && (
```
dawnho, line 33: This is meant to be for the whole section, not just the connect account row.

razor-x: This will hide the connected account info. The device id is not related to the connect account but may be hidden with `disableResourceIds`.

I feel the manufacturer information is fine to leave in here, but we could add a prop to hide that as well: `disableManufacturerInformation`.

PR #633, opened 2024-05-17
feat: Add noise sensor activity list

xplato:
Adds the activity event list to `NoiseSensorDeviceDetails`. This PR had to introduce several other components and functions, namely `TabSet` and `useEvents`.

## Demonstration

https://github.com/seamapi/react/assets/87919558/d242905a-5f89-4137-bc99-d860f7bf638b


src/lib/seam/components/DeviceDetails/DeviceDetails.stories.tsx
```diff
@@ -18,6 +18,7 @@ const meta: Meta<typeof DeviceDetails> = {
       type: 'figma',
       url: 'https://www.figma.com/file/Su3VO6yupz4yxe88fv0Uqa/Seam-Components?type=design&node-id=358-39439&mode=design&t=4OQwfRB8Mw8kT1rw-4',
     },
+    layout: 'fullscreen',
```
xplato, line 21: Storybook adds a default padding to previews, this removes it.

I wonder how many other renders aren't perfectly accurate due to this...

src/lib/ui/noise-sensor/NoiseSensorActivityList.tsx
```diff
@@ -0,0 +1,30 @@
+import type { NoiseSensorDevice } from 'seamapi'
+
+import { useEvents } from 'lib/seam/events/use-events.js'
+import { NoiseSensorEventItem } from 'lib/ui/noise-sensor/NoiseSensorEventItem.js'
+
+const since = new Date(Date.now() - 30 * 24 * 60 * 60 * 1000)
```
razor-x, line 6: This must `useCurrentTime`, see e.g., https://github.com/seamapi/react/pull/378#discussion_r1308080619

In this case, when should the result update? This implementation says "whenever the module is initialized" which is indeterminate behavior. The component should get the current time with `useCurrentTime` and then be explicit about when the time will update in the local component state (and thus refetch the query). 

What do you think is the right behavior here for the consumer?

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+function getContextLabel(event: Event): string {
+  switch (event.event_type) {
+    case 'noise_sensor.noise_threshold_triggered':
+      return `${t.noiseThresholdTriggered}`
+    case 'noise_detection.detected_noise':
+      return `${t.detectedNoise}`
+    default:
+      return t.noiseDetected
```
razor-x, line 49: When would this fallback ever happen?

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+
+function getContextSublabel(event: Event): string | null {
+  if (
+    'noise_threshold_name' in event &&
+    typeof event.noise_threshold_name === 'string'
```
razor-x, line 56: Prefer null checks, since otherwise this begs the question, "what other types can it be? A bool? A number?"

```suggestion
    typeof event.noise_threshold_name != null
```

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+  ) {
+    return event.noise_threshold_name
+  }
+
+  if (
+    'noise_level_decibels' in event &&
+    typeof event.noise_level_decibels === 'number'
```
razor-x, line 63:
```suggestion
    event.noise_level_decibels != null
```

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </div>
+  )
+}
+
+function getContextLabel(event: Event): string {
+  switch (event.event_type) {
+    case 'noise_sensor.noise_threshold_triggered':
+      return `${t.noiseThresholdTriggered}`
+    case 'noise_detection.detected_noise':
+      return `${t.detectedNoise}`
```
razor-x, line 47:
```suggestion
      return t.noiseThresholdTriggered
    case 'noise_detection.detected_noise':
      return t.detectedNoise
```

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <p className='seam-noise-sensor-event-item-date'>{date}</p>
+          <p className='seam-noise-sensor-event-item-time'>{time}</p>
+        </div>
+      </div>
+
+      <div className='seam-noise-sensor-event-item-context-wrap'>
+        <p className='seam-noise-sensor-event-item-context-label'>
+          {getContextLabel(event)}
+        </p>
+        {getContextSublabel(event) != null && (
```
razor-x, line 30: When would this be null?

src/lib/ui/TabSet.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }, [activeTab, tabs])
+
+  useLayoutEffect(() => {
+    calculateHighlightStyle()
+  }, [activeTab, calculateHighlightStyle])
+
+  useEffect(() => {
+    window.addEventListener('resize', calculateHighlightStyle)
+    return () => {
+      window.removeEventListener('resize', calculateHighlightStyle)
```
razor-x, line 49:
```suggestion
    globalThis.addEventListener?.('resize', calculateHighlightStyle)
    return () => {
      globalThis.removeEventListener?.('resize', calculateHighlightStyle)
```

src/lib/ui/TabSet.tsx
```diff
@@ -0,0 +1,104 @@
...
+  activeTab,
+  onTabChange,
+}: TabSetProps<TabType>): JSX.Element {
+  const [highlightStyle, setHighlightStyle] = useState<HighlightStyle>({
+    left: 0,
+    width: 140,
+  })
+
+  const calculateHighlightStyle = useCallback(() => {
+    const tabButton: HTMLButtonElement | null = document.querySelector(
```
razor-x, line 32:
```suggestion
    const tabButton: HTMLButtonElement | null = globalThis.document?.querySelector(
```

src/lib/seam/components/DeviceDetails/NoiseSensorDeviceDetails.tsx
```diff
@@ -22,38 +27,59 @@ export function NoiseSensorDeviceDetails({
   onBack,
   className,
 }: NoiseSensorDeviceDetailsProps): JSX.Element | null {
+  const [tab, setTab] = useState<TabType>('Details')
```
razor-x, line 30:
```suggestion
  const [tab, setTab] = useState<TabType>('details')
```

src/lib/seam/components/DeviceDetails/NoiseSensorDeviceDetails.tsx
```diff
@@ -8,7 +9,11 @@ import { DeviceImage } from 'lib/ui/device/DeviceImage.js'
 import { NoiseLevelStatus } from 'lib/ui/device/NoiseLevelStatus.js'
 import { OnlineStatus } from 'lib/ui/device/OnlineStatus.js'
 import { ContentHeader } from 'lib/ui/layout/ContentHeader.js'
+import { NoiseSensorActivityList } from 'lib/ui/noise-sensor/NoiseSensorActivityList.js'
 import { NoiseThresholdsList } from 'lib/ui/noise-sensor/NoiseThresholdsList.js'
+import { TabSet } from 'lib/ui/TabSet.js'
+
+type TabType = 'Details' | 'Activity'
```
razor-x, line 16: The tab type should not depend on the "text" rendered for the tab.

```suggestion
type TabType = 'details' | 'activity'
```

src/lib/seam/components/DeviceDetails/NoiseSensorDeviceDetails.tsx
```diff
@@ -22,38 +27,59 @@ export function NoiseSensorDeviceDetails({
...
+                  <NoiseLevelStatus device={device} />
+                  <DeviceModel device={device} />
+                </div>
               </div>
             </div>
           </div>
+
+          <TabSet<TabType>
+            tabs={['Details', 'Activity']}
+            activeTab={tab}
```
razor-x, line 61:
```suggestion
            tabs={['details', 'activity']}
            tabTitles={{
              details: t.details,
              activity: t.activity,
            }}
            activeTab={tab}
```

src/lib/dates.ts
```diff
@@ -44,3 +47,22 @@ export const parseDateTimePickerValue = (
   timeZone: string
 ): DateTime =>
   DateTime.fromISO(value).setZone(timeZone, { keepLocalTime: true })
+
+/**
+ * Formats a date time string to an object with date and time properties.
+ */
+export const formatDateTime = (
```
razor-x, line 54: Right now, [each component defines it's own date and time format locally](https://github.com/seamapi/react/blob/1221de8024782ab974d40a2a663bf870137c36c2/src/lib/seam/components/ClimateSettingScheduleTable/ClimateSettingScheduleRowDetails.tsx#L51-L55). If we could standardize all components to share the same format, then I agree we should have three functions in this file: `formatDate`, `formatTime`, an `formatDateTime`.

Since this is only used in one place, let's move it to the file that uses it (src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx), and simplify the usage

```ts
const formatDate = (dateTime: string) => DateTime.fromISO(dateTime).toFormat(ABBREVIATED_DATE_FORMAT)

const formatTime = (dateTime: string) => DateTime.fromISO(dateTime).toFormat(TIME_FORMAT)
```

src/lib/ui/noise-sensor/NoiseSensorActivityList.tsx
xplato: Good point! I'm not seeing the `useCurrentTime` hook in the repo currently. Was it removed at some point? I see it was implemented in #378 as mentioned.

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
xplato: Whoops, fragment of previous approach 😅

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
xplato: This was for TS satisfaction. I can adjust the approach if needed

src/lib/ui/noise-sensor/NoiseSensorActivityList.tsx
xplato: > What do you think is the right behavior here for the consumer?

Yeah, this is a good question. I mean, we only need to set the `since` parameter once when the component is first rendered, because all calls afterward will just pull from that range. I definitely need to set a `refetchInterval`, however. That could be 15-30 seconds or so for now.

xplato: If we're only setting the `since` param on initialization, is the usage of `useCurrentTime` necessary in that case?

src/lib/dates.ts
xplato: Gotcha, okay!

src/lib/ui/noise-sensor/NoiseSensorActivityList.tsx
razor-x: Ah, I guess it's called `useNow` https://github.com/seamapi/react/blob/main/src/lib/ui/use-now.ts

You can set a `refetchInterval`, but in order for the UI to load more events without losing the exiting ones, we should bind the time to the state so it's only updated on unmount / remount.

```ts
const since = useState(new Date())
```

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
razor-x: This should either fallback to null then, or the `Event` type should be narrowed.

I think this means we should provide a better `useEvents` hook, something that would narrow the type based on the `event_types` parameter. This is a good feature for the TS SDK, but we should be able to implement something here for now, even if it's just a generic param, like `useEvents<"noise_sensor.noise_threshold_triggered" | "noise_detection.detected_noise">`

xplato: I like the generic pattern on `useEvents` 👍. If it's cool with you, I'll move that to a subsequent PR to keep this one relatively isolated (also since the removal of the `detected_noise` event nullifies this specific issue).

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
```diff
@@ -1,16 +1,20 @@
+import { DateTime } from 'luxon'
 import type { Event } from 'seamapi'
 
-import { formatDateTime } from 'lib/dates.js'
 import { ClockIcon } from 'lib/icons/Clock.js'
 
+const ABBREVIATED_DATE_FORMAT = 'MMM d, yyyy'
+const TIME_FORMAT = 'h:mm a'
```
razor-x, line 7: Generally we [should avoid defining our own date format string](https://moment.github.io/luxon/#/formatting?id=consider-alternatives), instead use a built in format from luxon, or set it using this method: https://github.com/seamapi/react/blob/1221de8024782ab974d40a2a663bf870137c36c2/src/lib/seam/components/AccessCodeDetails/AccessCodeDetails.tsx#L247-L260

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
```diff
@@ -39,38 +43,25 @@ export function NoiseSensorEventItem({
...
-}
-
 function getContextSublabel(event: Event): string | null {
-  if (
-    'noise_threshold_name' in event &&
-    typeof event.noise_threshold_name === 'string'
-  ) {
-    return event.noise_threshold_name
+  if ('noise_threshold_name' in event) {
+    return `${event.noise_threshold_name}`
```
razor-x, line 48:
```suggestion
    return event.noise_threshold_name
```

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
xplato: Good point! I've replaced them with Luxon contacts.

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
xplato: This is due to the properties being `unknown` on the Event type (since payload appears to be poorly typed and perhaps incorrect in some places). Working on a reasonable solution...

xplato: Templating, while ugly, does work, however, as in the case below this.

razor-x: That's ok, we have better Event types coming! For now, you can add a `ts-expect-error` and file an `upstream` issue in this repo, e.g.,

```suggestion
    // @ts-expect-error <link to issue>
    return event.noise_threshold_name
```

Glad you pointed this out, since these look like minor "nits" but actually help us find gaps to ensure we don't return `undefined` in the UI :)

razor-x: Actually, just link https://github.com/seamapi/react/pull/633 :)

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
xplato: (Missed this earlier!) This is again from upstream event types. It's a fallback I wanted to get in so that, just in case those two properties are missing, the sub label UI is hidden and shows the label in the center.

src/lib/ui/noise-sensor/NoiseSensorActivityList.tsx
```diff
@@ -0,0 +1,31 @@
...
+  device,
+}: NoiseSensorActivityListProps): JSX.Element {
+  const date = useNow()
+  const [since] = useState(date)
+
+  const { events } = useEvents({
+    device_id: device.device_id,
+    event_type: 'noise_sensor.noise_threshold_triggered',
+    since: since.minus({ months: 1 }).toString(),
+  })
```
razor-x, line 22:
```suggestion
  const now = useNow()
  const [mountedAt] = useState(now)

  const { events } = useEvents({
    device_id: device.device_id,
    event_type: 'noise_sensor.noise_threshold_triggered',
    since: mountedAt.minus({ months: 1 }).toString(),
  })
```

razor-x (approved):

src/lib/ui/noise-sensor/NoiseSensorEventItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+function getContextSublabel(event: Event): string | null {
+  if ('noise_threshold_name' in event) {
+    // @ts-expect-error [upstream] shallow event type (#633)
+    return event.noise_threshold_name
+  }
+
+  if ('noise_level_decibels' in event) {
+    // Cast: [upstream] shallow event type (#633)
```
razor-x, line 50:
```suggestion
    // UPSTREAM: Shallow event types.
    // https://github.com/seamapi/react/issues/611
```

PR #644, opened 2024-07-18
Implement shake

razor-x:


src/lib/object.ts
```diff
@@ -0,0 +1,8 @@
+export const shake = <T>(obj: T): Partial<T> => {
```
phpnode, line 1:
```suggestion
type Shook<T> = {[K in keyof T as T[K] extends undefined ? never : K]: T[K]}
export const shake = <T>(obj: T): Shook<T> => {
```

phpnode (approved):

src/lib/object.ts
```diff
@@ -0,0 +1,13 @@
+export const shake = <T>(obj: T): Shook<T> => {
+  if (obj == null) return obj
+  if (typeof obj !== 'object') return obj
+
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+  const initialValue = {} as Shook<T>
```
razor-x, line 6: @phpnode I can't figure out how to do this without a type assertion. 🤷🏻

phpnode: you can't, TS isn't clever enough unfortunately.

phpnode: if we were being strict we'd say that this is a `Partial<T>` here, and it's only when we return that it's `as Shook<T>`, but I don't think it matters at all

Issue #646, opened 2024-07-22
update-devicedb-seed fails due to unexpected image url formats


louis-pre:
Running [update-devicedb-seed](https://github.com/seamapi/react/blob/main/.storybook/update-devicedb-seed.ts) fails because `seed-from-api` of the fake devicedb expects image urls to be formatted like this:

`https://connect.getseam.com/internal/devicedb_image_proxy?image_id=b132cbd2-33da-4806-bc33-a9273586faaa`

However, they are currently returned proxied by Next's image processing like:

`https://connect.getseam.com/_next/image?url=https://connect.getseam.com/internal/devicedb_image_proxy?image_id=b132cbd2-33da-4806-bc33-a9273586faaa&q=75&w=128`

Either we need to make the `seed-from-api` function more resilient to when image urls are proxied or we switch which api is called within `update-devicedb-seed` so that the images are not proxied (not sure if that's possible).

Stack trace:
```
➜  react git:(main) ✗ npm run storybook:update-devicedb-seed

> @seamapi/react@2.20.0 storybook:update-devicedb-seed
> tsx .storybook/update-devicedb-seed.ts

/Users/louispre/Workspace/react/node_modules/@seamapi/fake-devicedb/src/lib/database/seed-from-api.ts:105
  }
  ^


Error: No image_id in https://connect.getseam.com/_next/image?url=https://connect.getseam.com/internal/devicedb_image_proxy?image_id=682d539a-eca1-4a09-bf82-b2abfa8036ea&q=75&w=128
    at getImageIdFromImageUrl (/Users/louispre/Workspace/react/node_modules/@seamapi/fake-devicedb/src/lib/database/seed-from-api.ts:105:3)
    at seedDatabaseFromApi (/Users/louispre/Workspace/react/node_modules/@seamapi/fake-devicedb/src/lib/database/seed-from-api.ts:42:22)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at seedDatabase (/Users/louispre/Workspace/react/.storybook/update-devicedb-seed.ts:60:3)
    at main (/Users/louispre/Workspace/react/.storybook/update-devicedb-seed.ts:16:3)
    at <anonymous> (/Users/louispre/Workspace/react/.storybook/update-devicedb-seed.ts:63:1)
```

PR #648, opened 2024-07-22
test: Use version of fake-devicedb that supports proxied image urls

louis-pre:
https://github.com/seamapi/react/issues/646 was fixed upstream, we want to use the new version that contains the fix


razor-x:
Let's do all new PRs off of and back into the beta branch

razor-x (approved):

PR #656, opened 2024-08-26
fix: Unable to edit time_bound access codes

mastafit:


src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
```diff
@@ -173,7 +173,7 @@ function Content({
       : null
 
   const hasCodeInputs =
-    accessCode?.type !== 'time_bound' || accessCode.is_offline_access_code
+    accessCode?.type !== 'time_bound' && accessCode?.is_offline_access_code
```
razor-x, line 176: Won't this new condition mean that the UI will only allow editing offline access codes? What about editing online access codes?

mastafit: https://hello-seam.slack.com/archives/C04FS583FJA/p1724680131426929

src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
```diff
@@ -80,7 +80,7 @@ function Content({
     }
 
     const start = startDate.toISO()
-    if (start == null) {
+    if (start == null || start === undefined) {
```
razor-x, line 83: I am confused how this is not redundant 🤔 Wouldn't this already have be a linter error?
image

src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
razor-x: @mastafit I am looking at the thread, but it doesn't answer my original question. Can you answer my original question per the intent of this change?

mastafit: @razor-x You are right, I think we can add extra conditions

mastafit: I mean remove the extra conditions, something like that:
`const hasCodeInputs = accessCode?.type !== 'time_bound'`

razor-x: @mastafit This might help.

Code inputs were intentionally hidden for offline codes: https://github.com/seamapi/react/pull/606

But the current (and updated) logic also hides it for other code kinds. What if we only hide the inputs when for the offline code case?

```
const hasCodeInputs = accessCode.is_offline_access_code !== ture
```

src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
```diff
@@ -80,12 +80,12 @@ function Content({
     }
 
     const start = startDate.toISO()
-    if (start == null) {
+    if (start === null || start === undefined) {
       throw new Error(`Invalid start date: ${startDate.invalidReason}`)
     }
```
razor-x, line 85:
```suggestion
    if (start === null) {
      throw new Error(`Invalid start date: ${startDate.invalidReason}`)
    }
```

src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
```diff
@@ -80,12 +80,12 @@ function Content({
...
 
     const end = endDate.toISO()
-    if (end == null) {
+    if (end === null || end === undefined) {
```
razor-x, line 88:
```suggestion
    if (end === null) {
```

src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
```diff
@@ -206,81 +205,83 @@ function Content({
...
-                    </ul>
-                  )}
-                  <Button
-                    size='small'
-                    onMouseDown={(e) => {
-                      e.preventDefault() // Disable stealing input focus
-                      handleGenerateCode()
+          {hasCodeInputs !== null &&
+            hasCodeInputs !== undefined &&
+            hasCodeInputs && (
```
razor-x, line 210:
```suggestion
          {hasCodeInputs && (
```

src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
mastafit: Makes sense, I have updated the code!

src/lib/ui/AccessCodeForm/AccessCodeForm.tsx
mastafit: Yes, it's redundant, I have removed it now

razor-x (approved):

PR #659, opened 2024-10-08
Remove Climate Setting Schedules

### LINKED ISSUE #658: Fix compatibility with new thermostat types
Basic thermostat control should still work, but anything related to climate setting schedules will need to be removed and brought back later as climate presets plus thermostat schedules.

The most important thing is to fix any type errors with the latest types, as this will block users from updating.

mikewuu:
closes #658

### Notable things

- Removes ability to change default climate settings as it's been deprecated.


src/lib/seam/devices/use-devices.ts
```diff
@@ -1,13 +1,16 @@
...
+} from '@seamapi/http/connect'
 import { useQuery, useQueryClient } from '@tanstack/react-query'
 
 import { useSeamClient } from 'lib/seam/use-seam-client.js'
 import type { UseSeamQueryResult } from 'lib/seam/use-seam-query-result.js'
 
 export type UseDevicesParams = DevicesListParams
 
-export type UseDevicesData = Device[]
+export type UseDevicesData = DevicesListResponse['devices']
```
razor-x, line 13:
```suggestion
export type UseDevicesData = Device[]
```

src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
```diff
@@ -38,54 +33,15 @@ interface ThermostatDeviceDetailsProps
...
   device,
-  errorFilter,
-  warningFilter,
-  disableLockUnlock,
-  disableCreateAccessCode,
-  disableEditAccessCode,
-  disableDeleteAccessCode,
   disableResourceIds,
   disableConnectedAccountInformation,
-  disableClimateSettingSchedules,
```
razor-x, line 49: We should removed this from the API. Are there any other references to ClimateSettingSchedules in this project?

razor-x (changes requested):
I need to block merge because this will be a breaking change / new major version, and I don't want this to merge early.

src/lib/seam/components/DeviceDetails/ThermostatDeviceDetails.tsx
mikewuu: Yep. Removed all in https://github.com/seamapi/react/pull/659/commits/3457ca40076abd9cdd9c7840a6ec6c0502769d28

PR #670, opened 2025-01-20
feat: Add edit functionality to DeviceDetails

kadiryazici:
https://github.com/user-attachments/assets/53dedb39-5ec3-4a73-9c4a-26069f9788f1


src/lib/seam/components/SeamEditableDeviceName/SeamEditableDeviceName.tsx
```diff
@@ -0,0 +1,166 @@
+/* eslint-disable import/no-duplicates */
+
+import classNames from 'classnames'
+import type React from 'react'
+import { type KeyboardEvent, useCallback, useState } from 'react'
```
razor-x, line 5:
```suggestion
import classNames from 'classnames'
import type React from 'react'
import { type KeyboardEvent, type HTMLAttributes, type PropsWithChildren, type ChangeEvent, useCallback, useState } from 'react'
```

src/lib/seam/components/SeamEditableDeviceName/SeamEditableDeviceName.tsx
```diff
@@ -0,0 +1,166 @@
...
+  )
+}
+
+function fixName(name: string): string {
+  return name.replace(/\s+/g, ' ').trim()
+}
+
+function isValidName(
+  name: string
+): { type: 'success' } | { type: 'error'; message: string } {
```
razor-x, line 40: Kind of an arbitrary project convention, but we use `function` for components and arrow fns otherwise
```suggestion
const fixName = (name: string): string => {
  return name.replace(/\s+/g, ' ').trim()
}

const isValidName = (
  name: string
): { type: 'success' } | { type: 'error'; message: string } => {
```

src/lib/seam/components/SeamEditableDeviceName/SeamEditableDeviceName.tsx
```diff
@@ -0,0 +1,166 @@
...
+    setCurrentValue(fixedName)
+    onEdit?.(fixedName)
+  }, [currentValue, onEdit])
+
+  const handleChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>): void => {
+      setCurrentValue(event.target.value)
+      setErrorText(null)
+    },
+    []
```
razor-x, line 91: Why not this? Same question for some of the other `useCallback` dependency arrays below.
```suggestion
    [setCurrentValue, setErrorText]
```

src/lib/seam/components/SeamEditableDeviceName/SeamEditableDeviceName.tsx
```diff
@@ -0,0 +1,166 @@
...
+            </span>
+          )}
+        </span>
+      ) : (
+        <span>{value}</span>
+      )}
+
+      <span className='seam-editable-device-name-icon-wrapper'>
+        {editable ? (
+          editing ? (
```
razor-x, line 144: We really try to avoid nested ternaries. Normally we remove this by making an internal component and then use an early return, e.g.,


```ts
function SeamEditableDeviceName (...) {
   // hook stuff
   return (
     <Tag>
       <NameInput .../>
       <NameAction .../>
     </Tag>
   )
}

function NameInput (...) {
  if (editing) return <span>{value}</span>
  return (...)
}

function NameAction (...) {
  if (editable) return null
  if (editing) return (...)
  return (...)
}
```

src/lib/seam/devices/use-set-device-name.ts
```diff
@@ -0,0 +1,93 @@
...
+export type UseSetDeviceNameData = undefined
+
+export type UseSetDeviceNameMutationVariables = Pick<
+  DevicesUpdateBody,
+  'device_id' | 'name'
+>
+
+type MutationError = SeamHttpApiError
+
+export function useSetDeviceName(
```
razor-x, line 26: We like to provide the hooks and match the public API by default. Can we call this `useUpdateDevice`?

It can still only support updating the name (for now)

```
export type UseSetDeviceNameMutationVariables = Pick<
  DevicesUpdateBody,
  'device_id' | 'name'
>
```

razor-x (commented):
Great! Just a few minor comments on some non-obvious project conventions. I want make sure we test this against the API first before approving. Sounds like it works no problem on the fake.

src/lib/seam/components/SeamEditableDeviceName/SeamEditableDeviceName.tsx
kadiryazici: Because setters' references are always same but you're right, by following this scenario it's more strict.

razor-x (approved):

PR #671, opened 2025-01-22
feat: Improve icon scaling by using `viewBox` and `fontSize`

kadiryazici:
Without viewbox even if we resize SVG's their path's wouldn't resize. With this PR svgs are scaling according to `font-size`.
- Unnecessarry `scale: X` properties are deleted due to `font-size`


src/lib/ui/device/OnlineStatus.tsx
```diff
@@ -36,15 +36,15 @@ function AccountOnlineContent(props: { isOnline: boolean }): JSX.Element {
   if (isOnline) {
     return (
       <>
-        <OnlineStatusOnlineIcon />
+        <OnlineStatusOnlineIcon style={{ fontSize: '24px' }} />
```
razor-x, line 39: I was not 100% sure using `style` here is safe. Seam Components must support strict content security policy. Some `style` rules are ok to use because they use a different browser API (I cannot find the reference at the moment).

Anyway, I [tested this in the preview env](https://react-162orvajj.seam.vc/examples/basic/) to see if it triggers CSP errors and it seems ok. Maybe because it's a svg style? 

Just something to keep in mind when using `style` (we should always use css where possible, but I'm guessing this is not a case where it is possible).

kadiryazici: Oh  I'll create a className then. Style can be omitted

src/styles/_supported-device-table.scss
```diff
@@ -364,7 +365,7 @@ $see-all-button-height: 30px;
         }
       }
 
-      @media only screen and (width >= 450px) {
+      @media only screen and (width >=450px) {
```
razor-x, line 368:
```suggestion
      @media only screen and (width >= 450px) {
```

src/styles/_supported-device-table.scss
```diff
@@ -374,7 +375,7 @@ $see-all-button-height: 30px;
         }
       }
 
-      @media only screen and (width >= 768px) {
+      @media only screen and (width >=768px) {
```
razor-x, line 378:
```suggestion
      @media only screen and (width >= 768px) {
```

razor-x (approved):
@kadiryazici I undid some whitespace changes to trigger another deploy to get CI passing. I added you to vercel so future PR commits should work.

PR #674, opened 2025-01-29
fix: Hide pin code related fields if cannot specify

kadiryazici:


kadiryazici:
@razor-x I needed to update seaman package for types but now it's giving type error. I don't know what to do now 🤔

razor-x:
> @razor-x I needed to update seam/http package for types but now it's giving type error. I don't know what to do now 🤔 

You likely don't need to update that package. Try updating seamapi/types first instead.

src/lib/seam/components/AccessCodeDetails/AccessCodeDetails.tsx
```diff
@@ -94,6 +95,12 @@ export function AccessCodeDetails({
     )
   }, [accessCode, deleteCode, onDelete, preventDefaultOnDelete])
 
+  const { device } = useDevice({ device_id: accessCode?.device_id })
+  const cannotSpecifyPinCode =
+    device?.properties.code_constraints?.some(
+      ({ constraint_type: type }) => type === 'cannot_specify_pin_code'
+    ) ?? false
```
razor-x, line 102: Minor, but the double negative checks ("if not cannot foo...") are harder to grok
```suggestion
  const canSpecifyPinCode =
    (!device?.properties.code_constraints?.some(
      ({ constraint_type: type }) => type === 'cannot_specify_pin_code'
    )) ?? true
```

razor-x (approved):

src/lib/seam/components/AccessCodeDetails/AccessCodeDetails.tsx
kadiryazici: updated

razor-x (approved):

PR #675, opened 2025-01-29
fix: Icon size of access code table

kadiryazici:


src/lib/seam/components/AccessCodeTable/AccessCodeMainIcon.tsx
```diff
@@ -8,6 +8,10 @@ export interface AccessCodeMainIconProps {
   accessCode: AccessCode
 }
 
+function KeyIcon(): JSX.Element {
+  return <AccessCodeKeyIcon style={{ fontSize: '24px' }} />
+}
```
razor-x, line 13: This is another project convention: keep private functions below the main thing. So in this case, it would be moved to line 66.

razor-x (approved):

PR #678, opened 2025-01-29
feat: Add snackbars for toggling lock status

kadiryazici:


src/lib/seam/locks/use-toggle-lock-snackbar.tsx
```diff
@@ -0,0 +1,43 @@
...
+    ),
+    [visible, variant]
+  )
+
+  return {
+    showToggleSnackbar: useCallback((variant) => {
+      setVariant(variant)
+      setVisible(true)
+    }, []),
+    SnackbarNode,
```
mikewuu, line 36: If we are going to keep this snackbar hook, can we just rename this to `ToggleLockSnackbar` instead of the generic `SnackbarNode`?

src/lib/seam/locks/use-toggle-lock-snackbar.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const SnackbarNode = useMemo(
+    () => (
+      <Snackbar
+        variant={variant}
+        visible={visible}
+        onClose={() => {
+          setVisible(false)
+        }}
+        message={
+          variant === 'success' ? t.successfullyUpdated : t.failedToUpdate
```
mikewuu, line 23: Does this pattern mean we'll need a new hook (and file) for every single snackbar going forward? or maybe we only create one if it is being used in multiple places?

IMO duplicating seems simpler here instead of a separate hook.

mikewuu: @razor-x  any thoughts on this?

kadiryazici: the reason why I created a hook is there is two different components that are toggling snackbars. In order not to duplicate the code I created a hook for it.

mikewuu: Yea, I think we're ok to just duplicate it. Otherwise this pattern could result in too many hooks/files, for every snackbar type.

kadiryazici: @mikewuu updated it as using duplication

mikewuu (approved):

PR #679, opened 2025-01-29
feat: Update to @seamapi/http@v1.20.0

razor-x:
- **Update to @seamapi/types v1.344.1**
- **Update to @seamapi/http v1.20.0**


mikewuu:
This should fix the error https://github.com/seamapi/seam-connect/pull/10152

PR #715, opened 2025-11-13
Prefer null checks for capability flags

razor-x:


panzerchris (approved):
I didn't realise TS allowed any string for an in check. Kinda wild!

